PR #44, opened 2021-03-09
Update Reactr to support OnErr

cohix:


cohix:
Adding wait until there's a tagged version of Reactr to work off of.

PR #49, opened 2021-03-17
Clean command

samderlust:


cohix:
You've got the right idea here, but in this case we should use Go's built-in filesystem functions rather than `util.Run`. It makes sense to do that sometimes, but not in this case.

You can use `ioutil.ReadDir` to get a list of files in a directory, then for each one, you can check if the name matches one of the things you're looking for (and use `f.IsDir` to ensure the target and .build files are actually directories), and then use `os.RemoveAll` or `os.Remove` to delete things.

subo/command/clean.go
```diff
@@ -0,0 +1,67 @@
...
+			}
+
+			if len(bctx.Runnables) == 0 {
+				return errors.New("🚫 no runnables found in current directory (no .runnable yaml files found)")
+			}
+			logStart(fmt.Sprintf("cleaning in %s", bctx.Cwd))
+
+			//delete target or .build folder
+			dirs, _ := ioutil.ReadDir(".")
+			for _, dir := range dirs {
```
cohix, line 39: Since the `.target` and `build` dirs are inside the Runnable dirs, you'll need to run this once per Runnable in `bctx.Runnables`

test-project/fetch/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fetch"
-version = "0.1.0"
+version = "0.8.0"
```
cohix, line 3: Could you update these again to `0.9.0` ?

cohix (changes requested):

subo/command/clean.go
```diff
@@ -0,0 +1,64 @@
...
+				for _, dir := range dirs {
+					if dir.IsDir() {
+						if dir.Name() == "target" || dir.Name() == ".build" {
+							if rErr := os.RemoveAll(dir.Name()); rErr != nil {
+								logInfo(rErr.Error())
+							}
+							logDone(fmt.Sprintf("removed %s", dir.Name()))
+						}
+					} else {
+						if fileRegEx.MatchString(dir.Name()) {
```
cohix, line 51: Instead of RegEx, you can just check if `strings.HasSuffix` instead.

subo/command/clean.go
```diff
@@ -0,0 +1,64 @@
...
+
+			if len(bctx.Runnables) == 0 {
+				return errors.New("🚫 no runnables found in current directory (no .runnable yaml files found)")
+			}
+			logStart(fmt.Sprintf("cleaning in %s", bctx.Cwd))
+
+			//regex to find .wasm files
+			fileRegEx, _ := regexp.Compile("^.+\\.wasm")
+
+			for i := 0; i < len(bctx.Runnables); i++ {
```
cohix, line 39: You can do `for _, r := range bctx.Runnables` instead, and for each iteration of the loop, `r` will be one of the elements in the array.

subo/command/clean.go
```diff
@@ -0,0 +1,64 @@
...
+				//delete target or .build folder
+				dirs, _ := ioutil.ReadDir(".")
+				for _, dir := range dirs {
+					if dir.IsDir() {
+						if dir.Name() == "target" || dir.Name() == ".build" {
+							if rErr := os.RemoveAll(dir.Name()); rErr != nil {
+								logInfo(rErr.Error())
```
cohix, line 46: You should add a `continue` here so that it doesn't keep going and execute the `logDone`

subo/command/clean.go
```diff
@@ -0,0 +1,64 @@
...
+				return errors.New("🚫 no runnables found in current directory (no .runnable yaml files found)")
+			}
+			logStart(fmt.Sprintf("cleaning in %s", bctx.Cwd))
+
+			//regex to find .wasm files
+			fileRegEx, _ := regexp.Compile("^.+\\.wasm")
+
+			for i := 0; i < len(bctx.Runnables); i++ {
+				//delete target or .build folder
+				dirs, _ := ioutil.ReadDir(".")
```
cohix, line 41: You'll need to pass the Runnable dir's path instead of `.`

cohix (changes requested):

subo/command/clean.go
```diff
@@ -0,0 +1,62 @@
...
+					if dir.IsDir() {
+						if dir.Name() == "target" || dir.Name() == ".build" {
+							if rErr := os.RemoveAll(dir.Name()); rErr != nil {
+								logInfo(rErr.Error())
+							}
+							logDone(fmt.Sprintf("removed %s", dir.Name()))
+						}
+					} else {
+						if strings.HasSuffix(dir.Name(), ".wasm") {
+							os.Remove(dir.Name())
```
cohix, line 49: I ran this and it doesn't seem to be actually deleting things. You should check the error returned by `os.Remove`

cohix (approved):

Issue #31, opened 2021-01-27
Do not bundle when in a runnable directory


cohix:
When running `subo build` inside a Runnable directory, we should not create a bundle. as it doesn't make sense to create a bundle unless running from a project root.

PR #54, opened 2021-04-09
Return error if `subo build` is run inside a Runnables directory

Sylfrena:
Resolves https://github.com/suborbital/subo/issues/31

### DIFF
--- a/subo/command/build.go
+++ b/subo/command/build.go
@@ -2,6 +2,7 @@ package command
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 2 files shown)


Sylfrena:
@cohix  Please take a look, thanks!

subo/command/build.go
```diff
@@ -37,6 +38,15 @@ func BuildCmd() *cobra.Command {
 				return errors.New("🚫 no runnables found in current directory (no .runnable yaml files found)")
 			}
 
+			innerFiles, err := ioutil.ReadDir(bctx.Cwd)
+			if err != nil {
+				return errors.Wrapf(err, "Failed to list files in %s", bctx.Cwd)
+			}
+			fileName, _ := context.ContainsRunnableYaml(innerFiles)
+			if fileName == ".runnable.yaml" {
+				return errors.New("🚫 Cannot build inside a Runnable directory. Please run from project root.")
```
cohix, line 47: Instead of returning an error, let's just print a warning that says "building single Runnable, will not create a bundle" or something like that. Technically building just one Runnable is valid.

subo/command/build.go
```diff
@@ -37,6 +38,15 @@ func BuildCmd() *cobra.Command {
 				return errors.New("🚫 no runnables found in current directory (no .runnable yaml files found)")
 			}
 
+			innerFiles, err := ioutil.ReadDir(bctx.Cwd)
+			if err != nil {
+				return errors.Wrapf(err, "Failed to list files in %s", bctx.Cwd)
+			}
+			fileName, _ := context.ContainsRunnableYaml(innerFiles)
```
cohix, line 45: You can simplify this a bit by checking `if _, exists := context.ContainsRunnableYaml; exists { ... }`

cohix (changes requested):

cohix (approved):

cohix:
LGTM!

PR #66, opened 2021-06-12
make

samderlust:


subo/command/build.go
```diff
@@ -48,6 +48,14 @@ func BuildCmd() *cobra.Command {
 			shouldBundle := !noBundle && !bctx.CwdIsRunnable
 
 			useNative, _ := cmd.Flags().GetBool("native")
+			makeStr, _ := cmd.Flags().GetString("make")
```
cohix, line 51: Let's change the variable name to `makeTarget`

subo/command/build.go
```diff
@@ -48,6 +48,14 @@ func BuildCmd() *cobra.Command {
...
+
+			util.LogStart(fmt.Sprintf("make %s", makeStr))
+			_, _, err = util.Run(fmt.Sprintf("make %s", makeStr))
+
```
cohix, line 55: Please remove this blank line and add a blank line after the `if err != nil` closing brace.

cohix (changes requested):
Looking great!

subo/command/build.go
```diff
@@ -51,6 +51,14 @@ func BuildCmd() *cobra.Command {
 			shouldDockerBuild, _ := cmd.Flags().GetBool("docker")
 
 			useNative, _ := cmd.Flags().GetBool("native")
+			makeTarget, _ := cmd.Flags().GetString("make")
+
+			util.LogStart(fmt.Sprintf("make %s", makeTarget))
```
cohix, line 56: let's only run  this if `makeTarget` is not an empty string. Otherwise, looks good and I tested it!

cohix: (the whole block, i.e. only attempt to run make if the `makeTarget` isn't empty.

cohix (approved):

Issue #77, opened 2021-07-01
"failed to get CurrentBuildContext" when running "subo create project" in user home folder


FanaHOVA:
To reproduce:
- Install subo 
- `cd ~`
- `subo create project name`

Returns:
```
Error: 🚫 failed to get CurrentBuildContext: failed to getRunnableDirs: failed to list files in .Trash: open /Users/fana/.Trash: operation not permitted
```

Expected result:
Project is created successfully

Hardware specs:
- 2.8 GHz Quad-Core Intel Core i7
- 16 GB 2133 MHz LPDDR3
- Docker Desktop 3.5.1

cohix:
@FanaHOVA Thanks for filing.

The issue resides here: https://github.com/suborbital/subo/blob/main/builder/context/buildcontext.go#L117

When creating the `BuildContext`, the `getRunnableDirs` function attempts to look inside each directory to see if there's a `.runnable.yaml` file. If that lookup fails, we error the whole process. I'd suggest we just `continue` when that error occurs instead, which would just skip any directory that we don't have permission to read.

Let me know if you'd like to take this one on, otherwise I'll pick it up before the next release :)

PR #79, opened 2021-07-03
When unable to readDir in context, log error and continue

FanaHOVA:
Solves #77

### DIFF
--- a/builder/context/buildcontext.go
+++ b/builder/context/buildcontext.go
@@ -3,7 +3,6 @@ package context
 import (
 	"fmt"
 	"io/ioutil"
-	"log"
 	"os"
 	"path/filepath"
 	"strings"
@@ -11,6 +10,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/suborbital/atmo/directive"
 	"github.com/suborbital/subo/subo/release"
+	"github.com/suborbital/subo/subo/util"
 	"gopkg.in/yaml.v2"
 )
 
@@ -117,7 +117,7 @@ func getRunnableDirs(cwd string) ([]RunnableDir, bool, error) {
 		// determine if a .runnable file exists in that dir
 		innerFiles, err := ioutil.ReadDir(dirPath)
 		if err != nil {
-			log.Printf("couldn't read files in %v", dirPath)
+			util.LogWarn(fmt.Sprintf("couldn't read files in %v", dirPath))
 			continue
 		}
 
--- a/builder/context/buildcontext.go
+++ b/builder/context/buildcontext.go
@@ -3,6 +3,7 @@ package context
 import (
 	"fmt"
 	"io/ioutil"
+	"log"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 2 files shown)


builder/context/buildcontext.go
```diff
@@ -116,7 +117,8 @@ func getRunnableDirs(cwd string) ([]RunnableDir, bool, error) {
 		// determine if a .runnable file exists in that dir
 		innerFiles, err := ioutil.ReadDir(dirPath)
 		if err != nil {
-			return nil, false, errors.Wrapf(err, "failed to list files in %s", tf.Name())
+			log.Printf("couldn't read files in %v", dirPath)
```
cohix, line 120: There are some logging functions that most of `subo` uses (in the `util` package), could you create a `LogWarn` function in there and use it here, for consistency?

cohix (changes requested):

builder/context/buildcontext.go
FanaHOVA: @cohix Done, as a good Rubyist I've also DRY'd up the log util :)

subo/util/log.go
```diff
@@ -20,31 +21,36 @@ func (p *PrintLogger) LogInfo(msg string)  { LogInfo(msg) }
 func (p *PrintLogger) LogStart(msg string) { LogStart(msg) }
 func (p *PrintLogger) LogDone(msg string)  { LogDone(msg) }
 func (p *PrintLogger) LogFail(msg string)  { LogFail(msg) }
+func (p *PrintLogger) LogWarn(msg string)  { LogWarn(msg) }
 
-// LogInfo logs information
-func LogInfo(msg string) {
+// Keeping it DRY
+func Log(msg string) {
```
FanaHOVA, line 27: @cohix In Ruby I'd put this function in a `private` block since it's not supposed to be called from outside. I'm guessing here I'm doing the same thing by not adding `Log` to the `PrintLogger` struct above, is that correct? I want to make sure I'm doing it in the Go way :)

subo/util/log.go
```diff
@@ -20,31 +21,36 @@ func (p *PrintLogger) LogInfo(msg string)  { LogInfo(msg) }
...
 func LogFail(msg string) {
-	if _, exists := os.LookupEnv("SUBO_DOCKER"); !exists {
-		fmt.Println(fmt.Sprintf("🚫 FAILED: %s", msg))
-	}
+	Log(fmt.Sprintf("🚫 FAILED: %s", msg))
+}
+
+// LogWarn logs a warning from something
+func LogWarn(msg string) {
+	Log(fmt.Sprintf("⚠️ FAILED: %s", msg))
```
FanaHOVA, line 55: I tried inputting the warning emoji (⚠️) from both Chrome and VS Code, but it doesn't display it properly. Not sure why!

subo/util/log.go
cohix: Actually, you can make it private by just lowercasing it: `log(msg string)` :)

cohix (approved):

cohix:
Just the one change to unexport the `Log` function... otherwise looks great!

subo/util/log.go
cohix: It appears correctly for me :)

FanaHOVA:
@cohix Pretty neat that casing makes methods private/public! Just made the change. Let me know if I'm missing anything else!

cohix (approved):

PR #86, opened 2021-07-26
show help when type subo

samderlust:


jagger27:
Very weird behaviour from `spf13/cobra` here. Why does the root function run the function defined in the `create` `cobra.Command` struct? This patch does seem to work, though.

subo/root.go
```diff
@@ -26,6 +26,9 @@ Explore the available commands by running 'subo --help'`,
 		Use:   "create",
 		Short: "create a runnable or project",
 		Long:  `create a new Atmo project or WebAssembly runnable`,
+		RunE: func(cmd *cobra.Command, args []string) error {
```
cohix, line 29: Should this RunE be added to other top-level commands like `compute`, `create`, and `deploy` below?

subo/main.go
```diff
@@ -2,17 +2,11 @@ package main
...
 import (
 	"os"
-
-	"github.com/spf13/cobra"
 )
 
 func main() {
 	rootCmd := rootCommand()
 
-	rootCmd.RunE = func(cmd *cobra.Command, args []string) error {
```
cohix, line 12: You can set this field on the `cmd` object that's created inside the `rootCommand` function instead, right below the `Long` description.

subo/root.go
```diff
@@ -17,6 +17,12 @@ Subo is the full toolchain for using and managing Suborbital Development Platfor
 including building WebAssembly Runnables and Atmo projects.
 
 Explore the available commands by running 'subo --help'`,
+		RunE: func(cmd *cobra.Command, args []string) error {
```
jagger27, line 20: This makes much more sense.

subo/root.go
cohix: @samderlust are you adding it to the others too?

samderlust: yeah I can do that

cohix (approved):

jagger27 (changes requested):
I don't think we actually need to add these.
```go
if len(args) == 0 {
    cmd.Help()
}
return nil
```
Cobra provides per command help for free already, it seems. I built `main` from source and made sure I had the right binary and it already does the right thing:

```
$ ./subo compute                                            *[main]
manage Suborbital Compute Network resources

Usage:
  subo compute [command]

Available Commands:
  create      create compute network resources
  deploy      deploy compute network resources

Flags:
  -h, --help   help for compute

Use "subo compute [command] --help" for more information about a command.
```

jagger27:
A simpler fix is implemented in #89

Issue #109, opened 2021-10-21
CLI experience improvements


jagger27:
Remove the copy/paste step for getting an environment token by feeding it seamlessly through to the next command with a temporary file.

PR #111, opened 2021-10-26
feat: get token from tmp file

zhangwinning:
fix https://github.com/suborbital/subo/issues/109

cc @jagger27 @cohix

### DIFF
--- a/subo/command/compute_create_token.go
+++ b/subo/command/compute_create_token.go
@@ -35,7 +35,7 @@ func ComputeCreateTokenCommand() *cobra.Command {
 
 			fmt.Println(token.Token)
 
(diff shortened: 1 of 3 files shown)


cohix (approved):
@zhangwinning my apologies for letting this sit, I have been travelling.

This looks great to me. @jagger27 can you please test this locally and then merge if it all works as expected?

zhangwinning:
please cc @jagger27

jagger27 (changes requested):
Sorry for the long delay @zhangwinning, and thanks again for this contribution.

I'm wonder whether we should also remove the temporary file once it has been successfully copied to SCC.env.

Please rebase on `main` as well!

jagger27 (approved):

PR #113, opened 2021-11-04
Improvements to `subo build`

cohix:
This PR adds:

- A `--lang` flag to tell subo to only build certain languages
- Batched Docker builds; the builder images will now run just once each, and build all of the Runnables for their respective language instead of running once per Runnable
- Configurable mount path and relative build path for Docker Runnables


ospencer (approved):
Looks reasonable to me.

PR #114, opened 2021-11-09
Add local proxy and new function repl

cohix:
When running `subo compute deploy core --local`, subo will now create a proxy tunnel to the live `editor.suborbital.network` so that it runs on local HTTP (instead of HTTPS), which allows it to connect to the local HTTP editor (HTTPS pages can't access HTTP resources)

It also starts up a 'repl' that allows you to enter a function name, and it automatically creates an editor token and generates the local proxied editor URL for you.

Test command: `subo compute deploy core --local --branch connor/local-proxy --version dev`


MacklinEngineering (commented):
This looks good on my end, but will leave it to the go experts for a deep code dive.

subo/command/compute_deploy_core.go
```diff
@@ -117,6 +121,18 @@ func ComputeDeployCoreCommand() *cobra.Command {
...
 
+				proxy := localproxy.New("editor.suborbital.network")
+
+				go func() {
+					if err := proxy.Start(); err != nil {
+						log.Fatal(err)
+					}
+				}()
+
+				time.Sleep(time.Second * 2)
```
ospencer, line 132: I'm not a huge fan of this, but I do understand it's far less complex than using channels or something. Maybe add a comment?

ospencer (approved):

subo/command/compute_deploy_core.go
cohix: Sure thing

PR #115, opened 2021-11-12
Add goreleaser

### LINKED ISSUE #106: Add GitHub Actions for creating binary releases
Right now, we only provide Homebrew binaries on macOS. It would be useful to have pre-built binaries for other systems.

arbourd:
Adds an action that will update releases (when tags are created) uploading binaries and annotating the release notes with a `git diff`. If you don't want the diff, we can set the fetch depth back to 1 and just upload binaries.

Currently configured to build amd64/arm64 for Linux and macOS.

Closes #106

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,24 @@
+name: release
+
+on:
+  push:
+    tags:
+      - v*
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: 1.17
+
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v2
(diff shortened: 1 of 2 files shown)


arbourd:
An example with the same settings is ova here: https://github.com/arbourd/git-get/releases

cohix:
@arbourd I think I'd prefer to keep the release notes out of it since we have a process to add our own text to the release, so could we just stick with the binaries?

This is awesome, and so simple!

arbourd:
Sounds good! Any preference to the name of the checksums file (or including it at all)?

cohix:
@arbourd nope! LGTM :)

cohix (approved):

arbourd:
You might wanna give this a shot with a fake release (0.0.1-old) to ensure that the hooks are all gucci.

cohix:
@arbourd we test in prod 😈

PR #116, opened 2021-11-12
Add compiler flags

cohix:
This adds a "workaround" to support JSON in Assemblyscript by adding a compiler flag "analyzer" that automatically checks to see if the `--transform` flag is needed to build an AS module.

It's a bit hacky, but is the least bad way I could think to do it.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -3,6 +3,7 @@ package builder
 import (
 	"fmt"
 	"html/template"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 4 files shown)


builder/builder.go
```diff
@@ -271,3 +280,20 @@ func (b *Builder) checkAndRunPreReqs(runnable context.RunnableDir, result *Build
...
 	return nil
 }
+
+// analyzeForCompilerFlags looks at the Runnable and determines if any additional compiler flags are needed
+// this is initially added to support AS-JSON in AssemblyScript with its need for the --transform flag
+func (b *Builder) analyzeForCompilerFlags(runnable context.RunnableDir) (string, error) {
+	if runnable.Runnable.Lang == "assemblyscript" {
+		packageJSONBytes, err := ioutil.ReadFile(filepath.Join(runnable.Fullpath, "package.json"))
+		if err != nil {
+			return "", errors.Wrap(err, "faield to ReadFile package.json")
```
jagger27, line 290:
```suggestion
			return "", errors.Wrap(err, "failed to ReadFile package.json")
```
:^)

Issue #119, opened 2021-11-16
Subo 'create' command still makes a directory if it fails to find a template


jagger27:
Replication steps:

```
subo create runnable foobar --lang somethingnotsupported
```

Which still creates a `foobar` folder even though it failed.

denopink:
Got it replicated on my end. I'll try my hand at the patch.

denopink:
@jagger27 if `subo create runnable` fails, then wouldn't we want to delete the output dir all together (in this case we'll be deleting `foobar` and its contains) for any reason and not just for the current case `writeDotRunnable: somethingnotsupported is not an available language`?

jagger27:
> @jagger27 if `subo create runnable` fails, then wouldn't we want to delete the output dir all together (in this case we'll be deleting `foobar` and its contents) for any reason and not just for the current case `writeDotRunnable: somethingnotsupported is not an available language`?

Yes that seems reasonable.

PR #135, opened 2021-11-24
New Error and cleanup function for CreateRunnableCmd failures

denopink:
PR for #119 

Based on the conv in #119,  we want to delete the output dir of `sudo create runnable` for any failure that occurs in `CreateRunnableCmd`.

Replication steps:

```console
subo create runnable foobar --lang somethingnotsupported
```

`CreateRunnableError`s `Error` will also act as a cleanup function for these failures. atm, it's only deleting the output dir.

Ideally, cobra.Command would have a error callback feature when RunE fails... maybe we'll get one for Xmas. https://github.com/spf13/cobra/issues/914

Off topic, I wish cobra.Command had a error callback for each of their run `E` functions (`RunE`, `PreRunE`, `PostRunE`, `PersistentPreRunE`, `PersistentPostRunE`). They have a callback for when[ flag parsing fails](https://pkg.go.dev/github.com/spf13/cobra#Command.SetFlagErrorFunc). I'll create an issue ticket requesting this if it doesn't already exist.

### DIFF
--- a/subo/command/create_runnable.go
+++ b/subo/command/create_runnable.go
@@ -93,18 +93,18 @@ func CreateRunnableCmd() *cobra.Command {
 
 			runnable, err := writeDotRunnable(bctx.Cwd, name, lang, namespace)
 			if err != nil {
-				return errors.Wrap(CreateRunnableError{Path: path, Err: err}, "🚫 failed to writeDotRunnable")
+				return errors.Wrap(NewCreateRunnableError(path, err), "🚫 failed to writeDotRunnable")
 			}
 
 			templatesPath, err := template.TemplateFullPath(repo, branch)
 			if err != nil {
-				return errors.Wrap(CreateRunnableError{Path: path, Err: err}, "failed to TemplateDir")
+				return errors.Wrap(NewCreateRunnableError(path, err), "failed to TemplateDir")
 			}
 
 			if update, _ := cmd.Flags().GetBool(updateTemplatesFlag); update {
 				templatesPath, err = template.UpdateTemplates(repo, branch)
 				if err != nil {
-					return errors.Wrap(CreateRunnableError{Path: path, Err: err}, "🚫 failed to UpdateTemplates")
+					return errors.Wrap(NewCreateRunnableError(path, err), "🚫 failed to UpdateTemplates")
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


subo/command/create_runnable.go
```diff
@@ -45,6 +45,20 @@ var langAliases = map[string]string{
...
+	Err  error  // The original error.
+}
+
+// Error acts as a cleanup function for CreateRunnableError
+func (err CreateRunnableError) Error() string {
+	if cleanup_err := os.RemoveAll(err.Path); cleanup_err != nil {
+		fmt.Println("🚫 failed to clean up runnable outputs")
+	}
+	return err.Err.Error()
+}
```
jagger27, line 60: I'm not sure about having side effects in an `Error()` method. Maybe something like this instead?

We also shouldn't be printing here. We can just return the error.

```go
func NewCreateRunnableError(path string, err error) CreateRunnableError {
	if cleanupErr := os.RemoveAll(path); cleanupErr != nil {
		err = errors.Wrap(err, "failed to clean up runnable outputs")
	}
	return CreateRunnableError{Path: path, error: err}
}
```

subo/command/create_runnable.go
```diff
@@ -79,18 +93,18 @@ func CreateRunnableCmd() *cobra.Command {
 
 			runnable, err := writeDotRunnable(bctx.Cwd, name, lang, namespace)
 			if err != nil {
-				return errors.Wrap(err, "🚫 failed to writeDotRunnable")
+				return errors.Wrap(CreateRunnableError{Path: path, Err: err}, "🚫 failed to writeDotRunnable")
```
jagger27, line 96: These will need to be updated as well.

jagger27 (changes requested):
Some small suggestions. Let me know if you have any questions.

subo/command/create_runnable.go
```diff
@@ -45,6 +45,20 @@ var langAliases = map[string]string{
 	"gr":         "grain",
 }
 
+// Error for build command CreateRunnableCmd failures
+type CreateRunnableError struct {
+	Path string // The ouput directory for build command CreateRunnableCmd.
+	Err  error  // The original error.
+}
```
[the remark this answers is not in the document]
denopink: Ah, composition instead of inheritance! I haven't use this in go yet, pretty neat!

jagger27 (approved):
Looks good! Thanks!

jagger27:
Closes #119

cohix (approved):
Awesome!

Issue #126, opened 2021-11-19
Subo ignores --version flag when in existing compute installations 


jagger27:
If you have an existing Compute configuration setup: 
```
 subo compute deploy core --local
```

And you try to change the version by re-running the command in the same folder:
```
 subo compute deploy core --local --version v0.1.0-rc1
```
It seems to use the existing version.

denopink:
@jagger27 I was thinking we could also add the option of passing in the token as an env var  (maybe expecting it as a base64 encoded secret) for infra-as-code, playbook and debugging purposes? Should I open a separate ticket for this or included as a part of this patch?

denopink:
> @jagger27 I was thinking we could also add the option of passing in the token as an env var (maybe expecting it as a base64 encoded secret) for infra-as-code, playbook and debugging purposes? Should I open a separate ticket for this or included as a part of this patch?

Whoops, I missed `ReadEnvironmentToken`. Disregard my last comment!

denopink:
@jagger27 the issue appears to be that `docker-compose.yml` never updates

Replication steps:

**Fails to update (docker-compose.yml stays the same)**
```console
 subo compute deploy core --local
 subo compute deploy core --local --version v0.1.0-rc1
 ```
 **Update Succeeds (docker-compose.yml reflects changes)**
```console
 subo compute deploy core --local
 cd ../ && subo compute deploy core --local --version v0.1.0-rc1
 ```
The cause of the issue appears to be ExecTmplDir and how it skips any overwrites:
https://github.com/suborbital/subo/blob/750b9c98a15b54d74effdbf91bd0baf9c800a4af/builder/template/templates.go#L118-L128

A simple fix would do the trick and I'm assuming we don't always want to overwrite files, so we add a new overWrite bool parameter to ExecTmplDir aswell:
```go
func ExecTmplDir(cwd, name, templatesPath, tmplName string, templateData interface{}, overWrite bool) error {
	...
		// check if the target path is an existing file, and skip it if so unless overWrite is True
		if _, err := os.Stat(filepath.Join(targetPath, targetRelPath)); err != nil {
			if os.IsNotExist(err) {
				// that's fine, continue
			} else {
				return errors.Wrap(err, "failed to Stat")
			}
		} else if info.IsDir() || !overWrite {
			// if the target file already exists and overWrite is false, we're going to skip the rest since we don't want to overwrite
			return nil
		}
	...
```

cohix:
[...]

PR #138, opened 2021-11-24
Allow `subo compute deploy core --version [version]` to update deployment & Compute configuration

denopink:
PR for #126

Replication steps:

**Fails to update (docker-compose.yml stays the same)**
```console
 subo compute deploy core --local
 subo compute deploy core --local --version v0.1.0-rc1
 ```
 **Update Succeeds (docker-compose.yml reflects changes)**
```console
 subo compute deploy core --local
 cd ../ && subo compute deploy core --local --version v0.1.0-rc1
 ```
The cause of the issue appears to be ExecTmplDir and how it skips any overwrites:
https://github.com/suborbital/subo/blob/750b9c98a15b54d74effdbf91bd0baf9c800a4af/builder/template/templates.go#L118-L128

A simple fix would do the trick and I'm assuming we don't always want to overwrite files, so we should add a new overwrite bool parameter to ExecTmplDir aswell

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,7 @@ target
 
 # macOS
 .DS_Store
+
+# visual code
+.vscode/
+*.code-workspace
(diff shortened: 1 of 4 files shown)


denopink:
Already patched in `Subo CLI v0.2.1`. Closing PR.

PR #140, opened 2021-11-25
Add commit hash and build time to --version output

jagger27:
If for some reason Subo is built without the Makefile it falls back to the old output style. Timestamps are RFC3339 UTC format.

Example:
```
subo -v
Subo CLI v0.2.1 ad5be137b5b8dd68e2897459f5efa45d4534451c (Built at 2021-11-24T19:36:17Z)
```

### DIFF
--- a/subo/release/release.mk
+++ b/subo/release/release.mk
@@ -1,0 +1,5 @@
+now = $(shell date +'%Y-%m-%dT%TZ')
+commit = $(shell git rev-parse HEAD)
+var_path = github.com/suborbital/subo/subo/release
+RELEASE_FLAGS = "-X $(var_path).CommitHash=$(commit)\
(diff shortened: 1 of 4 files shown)


subo/release/release.mk
```diff
@@ -0,0 +1,5 @@
+now = $(shell date +'%Y-%m-%dT%TZ')
+commit = $(shell if [ ! -d .git ]; then echo "unknown"; else git rev-parse HEAD; fi)
```
cohix, line 2: That Git command outputs the "full" commit SHA, is there a way to use the shortened version (I think it's the prefix?)

cohix (commented):
This is great

cohix (approved):

Issue #139, opened 2021-11-24
Subo should check for new versions of itself


jagger27:
Subo is changing quickly. It would be useful `subo` could check for new versions of itself and notify the user that they should update.

denopink:
Two quick questions @jagger27:
1. do we want to add any version constraints since we're communicating with api.suborbital.network?
2. do we want to notify the user if they're using an unreleased version?

jagger27:
Resolved by #142

PR #142, opened 2021-11-28
Check for latest version of itself

denopink:
PR for #139

Check if the latest subo is running, otherwise notify the user that they should update.

### DIFF
--- a/root.go
+++ b/root.go
@@ -5,9 +5,14 @@ import (
 	"github.com/suborbital/subo/subo/command"
 	"github.com/suborbital/subo/subo/features"
 	"github.com/suborbital/subo/subo/release"
+	"github.com/suborbital/subo/subo/util"
 )
 
 func rootCommand() *cobra.Command {
+	err := release.CheckForLatestVersion()
+	if err != nil {
+		util.LogWarn(err.Error())
+	}
 	cmd := &cobra.Command{
 		Use:     "subo",
 		Short:   "Suborbital Development Platform CLI",
--- a/root.go
+++ b/root.go
@@ -9,10 +9,14 @@ import (
 )
 
 func rootCommand() *cobra.Command {
-	err := release.CheckForLatestVersion()
-	if err != nil {
-		util.LogWarn(err.Error())
-	}
+	defer func() {
+		version_msg, err := release.CheckForLatestVersion()
+		if err != nil {
+			util.LogFail(err.Error())
+		} else if version_msg != "" {
+			util.LogInfo(version_msg)
+		}
+	}()
 
 	cmd := &cobra.Command{
 		Use:     "subo",
--- a/root.go
+++ b/root.go
@@ -13,6 +13,7 @@ func rootCommand() *cobra.Command {
 	if err != nil {
 		util.LogWarn(err.Error())
 	}
+
 	cmd := &cobra.Command{
 		Use:     "subo",
 		Short:   "Suborbital Development Platform CLI",
--- a/root.go
+++ b/root.go
@@ -5,19 +5,9 @@ import (
 	"github.com/suborbital/subo/subo/command"
 	"github.com/suborbital/subo/subo/features"
 	"github.com/suborbital/subo/subo/release"
-	"github.com/suborbital/subo/subo/util"
 )
 
 func rootCommand() *cobra.Command {
-	defer func() {
-		version_msg, err := release.CheckForLatestVersion()
-		if err != nil {
-			util.LogFail(err.Error())
-		} else if version_msg != "" {
-			util.LogInfo(version_msg)
-		}
-	}()
-
 	cmd := &cobra.Command{
 		Use:     "subo",
 		Short:   "Suborbital Development Platform CLI",
--- a/subo/release/check.go
+++ b/subo/release/check.go
@@ -20,11 +20,11 @@ func getTimestampCache() (bool, error) {
 		return false, errors.Wrap(err, "failed to CacheDir")
 	}
 
-	var cached_timestamp time.Time
-	filePath := filepath.Join(cachePath, "latest_version_check.txt")
+	cachedTimestamp := time.Time{}
+	filePath := filepath.Join(cachePath, "subo_last_checked.txt")
 	_, err = os.Stat(filePath)
+
 	if os.IsNotExist(err) {
-		cached_timestamp = time.Time{}
 	} else if err != nil {
 		return false, errors.Wrap(err, "failed to Stat")
 	} else {
@@ -33,7 +33,7 @@ func getTimestampCache() (bool, error) {
 			return false, errors.Wrap(err, "failed to ReadFile")
 		}
 
-		cached_timestamp, err = time.Parse(time.RFC3339, string(data))
+		cachedTimestamp, err = time.Parse(time.RFC3339, string(data))
 		if err != nil {
 			return false, errors.Wrap(err, "failed to parse cached timestamp")
 		}
--- a/subo/release/check.go
+++ b/subo/release/check.go
@@ -113,12 +113,27 @@ func cacheLatestRelease(latestRepoRelease *github.RepositoryRelease) error {
 	return nil
 }
 
+func getLatestVersion() (*version.Version, error) {
+	latestRepoRelease, err := getLatestReleaseCache()
 	if err != nil {
-		return "", errors.Wrap(err, "failed to parse latest subo version")
+		return nil, errors.Wrap(err, "failed to getTimestampCache")
+	} else if latestRepoRelease == nil {
+		latestRepoRelease, _, err = github.NewClient(nil).Repositories.GetLatestRelease(context.Background(), "suborbital", "subo")
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to fetch latest subo release")
+		} else if err = cacheLatestRelease(latestRepoRelease); err != nil {
+			return nil, errors.Wrap(err, "failed to cacheLatestRelease")
+		}
 	}
 
-	cmdVersion, err := version.NewVersion(SuboDotVersion)
+	latestVersion, err := version.NewVersion(*latestRepoRelease.TagName)
 	if err != nil {
+		return nil, errors.Wrap(err, "failed to parse latest subo version")
+	}
+
+	return latestVersion, nil
+}
+
 		return "", errors.Wrap(err, "failed to parse current subo version")
 	} else if cmdVersion.LessThan(latestCmdVersion) {
 		return fmt.Sprintf("An upgrade for subo is available: %s → %s\n", cmdVersion, latestCmdVersion), nil
--- a/subo/release/check.go
+++ b/subo/release/check.go
@@ -77,7 +77,7 @@ func CheckForLatestVersion() (string, error) {
 	if err != nil {
 		return "", errors.Wrap(err, "failed to parse current subo version")
 	} else if cmdVersion.LessThan(latestCmdVersion) {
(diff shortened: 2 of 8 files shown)


subo/release/version.go
```diff
@@ -12,3 +21,22 @@ var AtmoVersion = "0.4.2"
 
 // SCCTag is the docker tag used for creating new compute core deployments
 var SCCTag = "v0.1.0"
+
+// CheckForLatestVersion returns an error if SuboDotVersion does not match the latest GitHub release or if the check fails
+func CheckForLatestVersion() error {
```
cohix, line 26: This is awesome! Can you add some spacing to your code in this function? We generally leave a blank line after each error check, for example. "Let it breathe!"

root.go
```diff
@@ -5,9 +5,14 @@ import (
...
 	"github.com/suborbital/subo/subo/features"
 	"github.com/suborbital/subo/subo/release"
+	"github.com/suborbital/subo/subo/util"
 )
 
 func rootCommand() *cobra.Command {
+	err := release.CheckForLatestVersion()
+	if err != nil {
+		util.LogWarn(err.Error())
+	}
```
cohix, line 15: Blank line after this pls :)

subo/release/version.go
```diff
@@ -12,3 +21,22 @@ var AtmoVersion = "0.4.2"
...
+		return errors.Wrap(err, "failed to fetch latest subo release")
+	}
+	latestCmdVersion, err := version.NewVersion(*latestRepoRelease.TagName)
+	if err != nil {
+		return errors.Wrap(err, "failed to parse latest subo version")
+	}
+	cmdVersion, err := version.NewVersion(SuboDotVersion)
+	if err != nil {
+		return errors.Wrap(err, "failed to parse current subo version")
+	} else if cmdVersion.LessThan(latestCmdVersion) {
```
cohix, line 38: I'm wondering if we should only print the warning when it's a new minor version? I wouldn't want to bug them for every patch version... what do you think @jagger27 ?

jagger27: I was working on this myself last week and I have a few different thoughts on it. 

We probably shouldn't be hammering the GitHub API every time subo runs. The check itself (as written above) takes on the order of 300ms on my connection and machine, which isn't great. I think we should also avoid doing an update check when subo is running in Docker, since it's not relevant there. I was thinking to `touch` a "last checked for updates" file in `~/.local/share/suborbital` and use that timestamp to check for updates less often.

Another idea I had was to defer the update notification to the termination of the program and spawn a goroutine to do the check in the background. 

I think using `util.LogInfo` would be more appropriate here as well, unless we start talking about patch severity like what `npm` does with some colourized output (https://github.com/fatih/color). It would also be useful to detect how subo was installed (Homebrew or not) and provide appropriate update instructions as well.

All of that is getting ahead of ourselves a bit, of course. :^)

```
⚠️ WARNING: upgrade subo 0.2.0 to the latest release 0.2.1
```
to
```
ℹ️  An upgrade for subo is available: v0.2.0 → v0.2.1
```

Personally, I want to be notified for 0.0.X releases, but I could see that getting annoying.

Other feedback, the update checker shouldn't consider an available update to be an error. So let's add a `bool` or maybe just the new version number as a `string` to the output of `CheckForLatestVersion()` along with any errors.

denopink: @jagger27 thx for the feedback! Tbh, I like both of your ideas to check a timestamp and to defer the check. If we don't defer the check, it'll be easy to miss the message.

So here's my notes so far:

1. Cache the latest version check timestamp in somewhere like `~/.local/share/suborbital` and use that timestamp to determine whether or not to check remotely again.
2. Defer the check
3. Switch from logging a warning to info
4. Update `CheckForLatestVersion()` output

jagger27: Yep, that's all good @denopink! Thanks for working on this.

It would also make sense to put all of the update related code in its own file (it can stay in the release package). `version.go` should stay as it was.

Also, make sure you're using the latest version of github.com/google/go-github.

denopink: Cool, just updated the code based on the feedback.

denopink:
@jagger27 so I completed that list, but atm we're still checking for versions when executing in a container. Are we thinking of using an env var to let subo know it's running in container or is there a cleaner why to check?

Also, do we want to create a new ticket for your other suggestions (they were pretty cool 🙂 )?
* detect how subo was installed (Homebrew or not) and provide appropriate update instructions
* add patch severity to the message (maybe we can add a json/yaml file in the release assets stating the severity)

root.go
```diff
@@ -5,9 +5,19 @@ import (
...
 
 func rootCommand() *cobra.Command {
+	defer func() {
+		version_msg, err := release.CheckForLatestVersion()
+		if err != nil {
+			util.LogFail(err.Error())
+		} else if version_msg != "" {
+			util.LogInfo(version_msg)
+		}
+	}()
```
jagger27, line 19: Let's move this into `main.go` since it's not part of the `cobra.Command` stuff. `defer` works fine here, but it doesn't make it concurrent. Something like this:

```go
func main() {
	rootCmd := rootCommand()

	done := update()
	if err := rootCmd.Execute(); err != nil {
		os.Exit(-1)
	}

	<-done
}

func update() chan bool {
	done := make(chan bool)

	go func() {
		version, err := release.CheckForLatestVersion()
		if err != nil {
			util.LogFail(err.Error())
		} else if version != "" {
			util.LogInfo(version)
		}
		done <- true
	}()

	return done
}
```

jagger27 (changes requested):
Awesome work @denopink! A few minor nitpicks and I think this will be good to go.

jagger27:
> @jagger27 so I completed that list, but atm we're still checking for versions when executing in a container. Are we thinking of using an env var to let subo know it's running in container or is there a cleaner why to check?
> 
> Also, do we want to create a new ticket for your other suggestions (they were pretty cool 🙂 )?
> 
>     * detect how subo was installed (Homebrew or not) and provide appropriate update instructions
> 
>     * add patch severity to the message (maybe we can add a json/yaml file in the release assets stating the severity)

Yes, please open new issues for those.

subo/release/check.go
```diff
@@ -0,0 +1,84 @@
...
+		cached_timestamp, err = time.Parse(time.RFC3339, string(data))
+		if err != nil {
+			return false, errors.Wrap(err, "failed to parse cached timestamp")
+		}
+	}
+
+	// check if 1 hour has passed since the last version check, and update the cached timestamp if so
+	current_timestamp := time.Now().UTC()
+	if cached_timestamp.IsZero() || current_timestamp.After(cached_timestamp.Add(time.Duration(1)*time.Hour)) {
+		data := []byte(current_timestamp.Format(time.RFC3339))
```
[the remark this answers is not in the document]
jagger27: Identifiers [should be likeThis or LikeThis](https://go.dev/doc/effective_go#mixed-caps).

subo/util/cache.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/pkg/errors"
+)
+
+// GetCacheDir returns and creates the cache directory
+func GetCacheDir() (string, error) {
+	cachePath, err := os.UserCacheDir()
```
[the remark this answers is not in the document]
cohix: @denopink this is sweet! We do however have an existing pattern of using $TMPDIR instead of the cache dir, so I think I'd prefer we stay consistent there... for example we could use `$TMPDIR/suborbital/cache`

See the `builder` package for examples where we use the tmpdir :)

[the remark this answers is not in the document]
jagger27: I've opened an issue on this: #143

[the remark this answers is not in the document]
denopink: Cool, I swapped out  `os.UserCacheDir()` with `os.TempDir()`

jagger27:
Just thought of one more thing while playing with this. If we previously found that an update is available, we should also cache that so that the update message is displayed again without having to check GitHub. It should still check if the timestamp is old.

denopink:
> Just thought of one more thing while playing with this. If we previously found that an update is available, we should also cache that so that the update message is displayed again without having to check GitHub. It should still check if the timestamp is old.

@jagger27 Cool, I just updated the code to reflect the release data caching.

subo/release/check.go
```diff
@@ -14,67 +16,130 @@ import (
...
+		return time.Time{}, errors.Wrap(err, "failed to CacheDir")
 	}
 
-	var cached_timestamp time.Time
-	filePath := filepath.Join(cachePath, "latest_version_check.txt")
-	_, err = os.Stat(filePath)
-	if os.IsNotExist(err) {
-		cached_timestamp = time.Time{}
+	cachedTimestamp := time.Time{}
+	filePath := filepath.Join(cachePath, "subo_last_checked.txt")
```
jagger27, line 26:
```suggestion
	filePath := filepath.Join(cachePath, "subo_last_checked")
```

subo/util/cache.go
```diff
@@ -7,14 +7,9 @@ import (
...
-func GetCacheDir() (string, error) {
-	cachePath, err := os.UserCacheDir()
-	if err != nil {
-		return "", errors.Wrap(err, "failed to get UserCacheDir")
-	}
-
-	targetPath := filepath.Join(cachePath, "suborbital")
+// CacheDir returns the cache directory and creates it if it doesn't exist
+func CacheDir() (string, error) {
+	targetPath := filepath.Join(os.TempDir(), "suborbital", "cache")
```
jagger27, line 12:
```suggestion
	targetPath := filepath.Join(os.TempDir(), "suborbital", "subo")
```

subo/release/check.go
```diff
@@ -14,67 +16,130 @@ import (
 	"github.com/suborbital/subo/subo/util"
 )
 
-func getTimestampCache() (bool, error) {
-	cachePath, err := util.GetCacheDir()
+func getTimestampCache() (time.Time, error) {
```
jagger27, line 19:
```suggestion
const lastCheckedFilename = "subo_last_checked"

func getTimestampCache() (time.Time, error) {
```

subo/release/check.go
jagger27, line 26:
```suggestion
	filePath := filepath.Join(cachePath, lastCheckedFilename)
```

subo/release/check.go
```diff
@@ -14,67 +16,130 @@ import (
...
-		data := []byte(current_timestamp.Format(time.RFC3339))
-		if err := ioutil.WriteFile(filePath, data, os.ModePerm); err != nil {
-			return false, errors.Wrap(err, "failed to WriteFile")
+func cacheTimestamp(timestamp time.Time) error {
+	cachePath, err := util.CacheDir()
+	if err != nil {
+		return errors.Wrap(err, "failed to CacheDir")
+	}
+
+	filePath := filepath.Join(cachePath, "subo_last_checked.txt")
```
jagger27, line 50:
```suggestion
	filePath := filepath.Join(cachePath, lastCheckedFilename)
```

So we don't have the same hardcoded string in two places.

jagger27 (approved):
Looking good. Just a few minor nitpicks and then we'll get someone else's eyes on it. Thank you Owen!

jagger27:
I'm pretty sure the `<-done` won't ever get stuck forever, but just to be safe we might want to add our own timeout.

denopink:
@jagger27 thx again for all the feedback as always! I just added the timeout and cleaned up a little bit.

cohix (approved):

Issue #147, opened 2021-12-05
Detect how subo was installed


denopink:
Detect how subo was installed (Homebrew or not) so subo can provide appropriate update instructions (if an update is available).

DevChoganwala:
@denopink, `brew list <formula>` will exit with exit code 1 if no formula for the specified package is found. We can use `exec` to execute this and detect if subo was installed using Homebrew.
I would like to work on this!!
Also, we would want to use this logic in `check.go` file that has not been merged with the main repo yet right?

denopink:
@DevChoganwala yup, that's how I would have done it imo.
Based on @jagger27 feedback, it'll be added to CheckForLatestVersion()'s message. So let's throw it in check.go.

jagger27:
> @denopink, `brew list <formula>` will exit with exit code 1 if no formula for the specified package is found. We can use `exec` to execute this and detect if subo was installed using Homebrew. I would like to work on this!! Also, we would want to use this logic in `check.go` file that has not been merged with the main repo or in the `version.go` file?

That's clever. I like that!

DevChoganwala:
Just so I am clear about this, there appear to be three main ways to get subo:
1. Through Homebrew (can be updated using `brew upgrade subo`)
2. Cloning this repository (can be updated using `git pull`)
3. Downloading source code manually (can be updated by going to https://github.com/suborbital/subo/releases)

We'll put logic for checking this in a separate file and use this in `check.go`

jagger27:
> Just so I am clear about this, there appear to be three main ways to get subo:
> 
>     1. Through Homebrew (can be updated using `brew upgrade subo`)
> 
>     2. Cloning this repository (can be updated using `git pull`)
> 
>     3. Downloading source code manually (can be updated by going to https://github.com/suborbital/subo/releases)
> 
> 
> We'll put logic for checking this in a separate file and use this in `check.go`

Yes, Homebrew and building from source are options, and we also provide binary builds. The key will be detecting if Subo was built from git HEAD or some tagged release.

DevChoganwala:
> > Just so I am clear about this, there appear to be three main ways to get subo:
> > ```
> > 1. Through Homebrew (can be updated using `brew upgrade subo`)
> > 
> > 2. Cloning this repository (can be updated using `git pull`)
> > 
> > 3. Downloading source code manually (can be updated by going to https://github.com/suborbital/subo/releases)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
[...]

PR #150, opened 2021-12-09
Add subo install method check

DevChoganwala:
Checks whether subo was installed using brew, cloned or built manually using source code.
Pull Request for issue #147

### DIFF
--- a/subo/release/check.go
+++ b/subo/release/check.go
@@ -145,11 +145,7 @@ func CheckForLatestVersion() (string, error) {
 	} else if cmdVersion, err := version.NewVersion(SuboDotVersion); err != nil {
 		return "", errors.Wrap(err, "failed to parse current subo version")
 	} else if cmdVersion.LessThan(latestCmdVersion) {
-		check, err := InstallInfo(latestCmdVersion)
-		if err != nil {
-			return fmt.Sprintf("An upgrade for subo is available: %s → %s\n", cmdVersion, latestCmdVersion), nil
-		}
-		return fmt.Sprintf("An upgrade for subo is available: %s → %s\n%s", cmdVersion, latestCmdVersion, check), nil
+		return fmt.Sprintf("An upgrade for subo is available: %s → %s\n%s", cmdVersion, latestCmdVersion, InstallInfo(latestCmdVersion)), nil
 	}
 
 	return "", nil
--- a/subo/release/check.go
+++ b/subo/release/check.go
@@ -145,7 +145,11 @@ func CheckForLatestVersion() (string, error) {
 	} else if cmdVersion, err := version.NewVersion(SuboDotVersion); err != nil {
 		return "", errors.Wrap(err, "failed to parse current subo version")
 	} else if cmdVersion.LessThan(latestCmdVersion) {
-		return fmt.Sprintf("An upgrade for subo is available: %s → %s\n", cmdVersion, latestCmdVersion), nil
+		check, err := InstallInfo(latestCmdVersion)
+		if err != nil {
+			return fmt.Sprintf("An upgrade for subo is available: %s → %s\n", cmdVersion, latestCmdVersion), nil
+		}
+		return fmt.Sprintf("An upgrade for subo is available: %s → %s\n%s", cmdVersion, latestCmdVersion, check), nil
 	}
 
 	return "", nil
--- a/subo/release/check.go
+++ b/subo/release/check.go
@@ -145,7 +145,7 @@ func CheckForLatestVersion() (string, error) {
 	} else if cmdVersion, err := version.NewVersion(SuboDotVersion); err != nil {
 		return "", errors.Wrap(err, "failed to parse current subo version")
 	} else if cmdVersion.LessThan(latestCmdVersion) {
-		return fmt.Sprintf("An upgrade for subo is available: %s → %s\n%s", cmdVersion, latestCmdVersion, InstallInfo(latestCmdVersion)), nil
+		return fmt.Sprintf("An upgrade for subo is available: %s → %s\n%s", cmdVersion, latestCmdVersion, InstallInfo()), nil
 	}
 
 	return "", nil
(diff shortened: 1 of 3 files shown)


subo/release/check_install.go
```diff
@@ -0,0 +1,57 @@
...
+
+		if err != nil {
+			return 2, "", nil
+		}
+
+		if strings.TrimRight(string(cmd), "\n") == remote {
+			return 1, mydir, nil
+		}
+
+		return 2, "", nil
```
Mithil467, line 53: This might be better written as -
```go
if err == nil {
if strings.TrimRight(string(cmd), "\n") == remote {
	return 1, mydir, nil
}
return 2, "", nil
```
Or logical AND can be used instead of 2 `if` checks.

Mithil467 (changes requested):
I am new here and thought would leave a general review. Some of these changes might not be relevant anymore if we consider the new comments under the issue.

DevChoganwala:
@denopink @Mithil467 , made changes according to the discussion and reviews. PTAL :)

subo/release/check_install.go
```diff
@@ -0,0 +1,20 @@
+package release
+
+import (
+	"fmt"
+	"os/exec"
+
+	"github.com/hashicorp/go-version"
+)
+
+func InstallInfo(v *version.Version) string {
```
denopink, line 10: Since we agreed to point the user to https://github.com/suborbital/subo#installing, we can remove the version reference 
```suggestion
func InstallInfo() string {
```

subo/release/check_install.go
```diff
@@ -0,0 +1,20 @@
...
+
+	_, err := exec.Command("brew", "list", "subo").Output()
+
+	if err != nil {
+		return fmt.Sprintf("Head over to https://github.com/suborbital/subo/releases/tag/v%s or pull new changes", v)
```
denopink, line 15: Update message to point user to https://github.com/suborbital/subo#installing

subo/release/check_install.go
```diff
@@ -0,0 +1,20 @@
...
+	}
+
+	return "Run: 'brew upgrade subo' to update subo"
+
+}
```
denopink, line 20: Remove newline
```suggestion
	return "Run: 'brew upgrade subo' to update subo"
}
```

subo/release/check.go
```diff
@@ -145,7 +145,7 @@ func CheckForLatestVersion() (string, error) {
 	} else if cmdVersion, err := version.NewVersion(SuboDotVersion); err != nil {
 		return "", errors.Wrap(err, "failed to parse current subo version")
 	} else if cmdVersion.LessThan(latestCmdVersion) {
-		return fmt.Sprintf("An upgrade for subo is available: %s → %s\n", cmdVersion, latestCmdVersion), nil
+		return fmt.Sprintf("An upgrade for subo is available: %s → %s\n%s", cmdVersion, latestCmdVersion, InstallInfo(latestCmdVersion)), nil
```
denopink, line 148: Based on my other comment
```suggestion
		return fmt.Sprintf("An upgrade for subo is available: %s → %s\n%s", cmdVersion, latestCmdVersion, InstallInfo()), nil
```

denopink (changes requested):
Looks good, I had some minor nitpicks. Bench marked `exec.Command("brew", "list", "subo").Output()` and it'll costs us around 1.5-ish seconds.

The only time I think users will notice is when they run something like `subo -v`, otherwise it'll probably complete before any other cmd.

@jagger27 would we feel comfortable adding that addition IO?

denopink:
Closes #147

subo/release/check_install.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"os/exec"
+)
+
+// InstallInfo checks if subo was installed using Homebrew or via some other method and returns update instructions
+func InstallInfo() string {
+	if _, err := exec.Command("brew", "list", "subo").Output(); err != nil {
+		return "Check out our install/upgrade instructions at https://github.com/suborbital/subo#installing"
+	}
+	return "Run: 'brew upgrade subo' to update subo"
```
denopink, line 12:
```suggestion
	}
	
	return "Run: 'brew upgrade subo' to update subo"
```

Mithil467 (approved):
One thing to be considered before merging might be the suggestion by @denopink to just check OS instead of running `brew list`. Otherwise, lgtm.

jagger27:
> One thing to be considered before merging might be the suggestion by @denopink to just check OS instead of running `brew list`. Otherwise, lgtm.

Although most macOS devs use Homebrew, there are definitely some who don't. Testing this out I found that the 2500 ms timeout wasn't long enough on my machine, probably because of how many Homebrew packages I have installed.

I think a better approach would be a compile-time flag for different "releases" of Subo. See [this PR](https://github.com/suborbital/subo/pull/140/files#diff-95bd414812289401a96a8ce0615eb90d545c36b4d5663d3a0445d8008437fcdc) for one technique.

Ideally there would be a variable indicating whether Subo was installed from:

- Homebrew
- GitHub action goreleaser tarball
- built from source
- built from source for use within Docker containers (so we can skip update checks and so on)
- and any future builds

denopink:
@jagger27 that is so cool! I have to read up more on `make`.

denopink (changes requested):
I really like @jagger27 suggestion https://github.com/suborbital/subo/pull/150#issuecomment-994971486 over `brew list` and os check.

subo/release/check_install.go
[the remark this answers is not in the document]
denopink: ^--- ignore since https://github.com/suborbital/subo/pull/150#issuecomment-994971486 is better

cohix:
@DevChoganwala I really appreciate you taking the time to try out this method, but It seems that the consensus is to fix this with build-time variables rather than the brew command. I am going to close this, but please let me know if you have any questions!

Issue #152, opened 2021-12-10
`subo compute deploy core` removes changes to `docker-compose.yml`


ospencer:
Compute users will change the `docker-compose.yml` file to update various configurations, like `SCC_TEMPLATES_REPO`. If they run `subo compute deploy core --local` again, it regenerates `docker-compose.yml`.

This was done because the `--version` flag would not update the images referenced in `docker-compose.yml`. We'll need to be a bit smarter and update relevant fields instead.

PR #154, opened 2021-12-12
Add --reset flag to compute deploy core command

cohix:
This adds this to `subo compute deploy core`:

--reset: resets the `docker-compose.yaml` or K8s manifests to the default template (does not update templates)

The reset behaviour is disabled unless the flag is passed, which is a breaking change.

If there are no manifests in the current directory, it will create them. If no templates exist, the latest are downloaded.

Resolves https://github.com/suborbital/subo/issues/152

### DIFF
--- a/builder/template/templates.go
+++ b/builder/template/templates.go
@@ -57,6 +57,34 @@ func UpdateTemplates(repo, branch string) (string, error) {
 	return tmplPath, nil
 }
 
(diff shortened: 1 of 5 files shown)


cohix:
Resolves https://github.com/suborbital/subo/issues/152

ospencer (approved):
Looks great!

PR #156, opened 2021-12-13
TinyGo "slim" and builder images

jagger27:
This PR splits the previous TinyGo builder image into two separate images: one that builds TinyGo (and its huge dependencies) and one for actually running TinyGo.


builder/builder.mk
```diff
@@ -38,13 +38,16 @@ builder/docker/swift/dev/publish:
 	docker buildx build . -f builder/docker/swift/Dockerfile --platform linux/amd64,linux/arm64 -t suborbital/builder-swift:dev --push
 
 # tinygo docker targets
-builder/docker/tinygo:
+builder/docker/tinygo-builder:
```
cohix, line 41: Let's call this `tinygo-base`

cohix: (the make target and the actual image)

builder/docker/tinygo/Dockerfile
```diff
@@ -1,23 +1,17 @@
...
+WORKDIR /usr/local
+RUN tar -xzf release.tar.gz \
+    && rm release.tar.gz
+
+RUN ln -s /usr/local/tinygo/bin/tinygo /usr/local/bin/tinygo
+RUN ln -s /usr/local/tinygo/bin/wasm-opt /usr/local/bin/wasm-opt
 
 WORKDIR /root/runnable
+
+RUN go mod download github.com/suborbital/reactr@v0.13
```
cohix, line 17: If we omit the version, won't it just always download the latest?

builder/builder.mk
```diff
@@ -38,13 +38,16 @@ builder/docker/swift/dev/publish:
...
+	docker build . -f builder/docker/tinygo/Dockerfile.builder -t suborbital/builder-tinygo-builder:$(ver)
+
+builder/docker/tinygo: builder/docker/tinygo-builder
```
cohix, line 44: I don't particularly want to build the base image every time... could we instead tag the base image based on the version of TinyGo, and just push it to Dockerhub?

cohix (approved):

Issue #168, opened 2022-01-07
version check continuously fails if cached timestamp is an empty string


denopink:
We discovered it during a code review for #22 .

The error was caused by an empty string saved as the cached timestamp.
`ℹ️  failed to getLatestVersion: failed to getTimestampCache: failed to getTimestampCache: failed to parse cached timestamp: parsing time "" as "2006-01-02T15:04:05Z07:00": cannot parse "" as "2006"`

Oddly enough, I have no idea how this occurred... since an empty timestamp is always converted to `0001-01-01T00:00:00Z` depending on the layout arg.

However, subo should just purge the bad data. Currently it does not do that and it just continuously fails due to the empty string.

Might as well do this for the cached release data.

cohix:
Done! Thanks @denopink

PR #169, opened 2022-01-08
bug patch: purge bad data cache from version check

denopink:
PR for #168

The bug is caused when an empty string is saved as the cached timestamp for `version check`.
ℹ️ failed to getLatestVersion: failed to getTimestampCache: failed to getTimestampCache: failed to parse cached timestamp: parsing time "" as "2006-01-02T15:04:05Z07:00": cannot parse "" as "2006"

Until this patch, subo did not purge the bad cache for `version check` and so `version check` will continuously fail. Note, `version check` ran in a separate thread so it didn't affect any executed command.

We also applied the same patch for `version check`'s cached repo release data as well. Not because we saw the same error for it, but it "could" happen in the future if a bad repo release data was cached some how.

Cheers! 🍻


denopink:
Wrong branch! Closing and opening a new pr.

PR #170, opened 2022-01-08
 bug patch: purge bad data cache from version check

denopink:
PR for #168

The bug is caused when an empty string is saved as the cached timestamp for `version check`.
ℹ️ failed to getLatestVersion: failed to getTimestampCache: failed to getTimestampCache: failed to parse cached timestamp: parsing time "" as "2006-01-02T15:04:05Z07:00": cannot parse "" as "2006"

Until this patch, subo did not purge the bad cache for `version check` and so `version check` will continuously fail. Note, `version check` ran in a separate thread so it didn't affect any executed command.

We also applied the same patch for `version check`'s cached repo release data as well. Not because we saw the same error for it, but it "could" happen in the future if a bad repo release data was cached some how.

Cheers! 🍻

### DIFF
--- a/subo/release/check.go
+++ b/subo/release/check.go
@@ -99,7 +99,7 @@ func getLatestReleaseCache() (*github.RepositoryRelease, error) {
 		if err != nil {
 			err_remove := os.Remove(filePath)
 			if err_remove != nil {
-				return nil, errors.Wrap(err, "faild to Remove bad cached repo release")
+				return nil, errors.Wrap(err, "failed to Remove bad cached RepositoryRelease")
 			}
 			return nil, errors.Wrap(err, "failed to Decode cached RepositoryRelease")
 		}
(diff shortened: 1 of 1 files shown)


subo/release/check.go
```diff
@@ -38,6 +38,10 @@ func getTimestampCache() (time.Time, error) {
 
 		cachedTimestamp, err = time.Parse(time.RFC3339, string(data))
 		if err != nil {
+			err_remove := os.Remove(filePath)
+			if err_remove != nil {
+				return time.Time{}, errors.Wrap(err, "faild to Remove bad cached timestamp")
```
jagger27, line 43:
```suggestion
			errRemove := os.Remove(filePath)
			if errRemove != nil {
				return time.Time{}, errors.Wrap(err, "failed to Remove bad cached timestamp")
```

jagger27 (changes requested):
Looks good, thank you @denopink!

jagger27 (approved):

PR #172, opened 2022-01-13
Reduce file and folder permissions across subo

### LINKED ISSUE #171: subo create runnable creates files with exec permissions
On Linux (Ubuntu 20.04, bash) `subo create runnable foo` creates a Rust runnable with the files having `+x` (execute) bit set on them which seems unneccessary.
image

jagger27:
I went through Subo and adjusted the file permissions according to the following rules:

- directories are all `0755`
- locally cached and temporary files are `0644`
- runnable source files are `0600`

Fixes #171

### DIFF
--- a/builder/context/directive.go
+++ b/builder/context/directive.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/suborbital/atmo/directive"
+	"github.com/suborbital/subo/subo/util"
 	"gopkg.in/yaml.v2"
 )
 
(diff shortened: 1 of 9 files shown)


denopink (commented):
Would it help if made these permissions constants in `package util`? It may help future contributors if we had something as obvious as `util.FilePrem` and etc.

cohix:
@jagger27 I like @denopink 's suggestion :) Let's do that, and then this looks good!

denopink (approved):
LGTM 🍻

PR #173, opened 2022-01-19
feat(suboCLI): create_handler command

MacklinEngineering:


subo/command/create_handler.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func CreateHandlerCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "handler <name>",
+		Short: "create a new handler",
+		Long:  `create a new handler for Subo CLI`,
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			name := args[0]
```
cohix, line 26: I was actually expecting `resource` to be the argument, i.e. `subo create handler /foo`. Could you explain why `name` was chosen?

subo/command/create_handler.go
```diff
@@ -0,0 +1,73 @@
...
+
+			util.LogStart(fmt.Sprintf("creating handler with function name %s", name))
+
+			bctx, err := context.ForDirectory(dir)
+			if err != nil {
+				return errors.Wrap(err, "failed to ForDirectory")
+			}
+
+			if bctx.Directive == nil {
+				util.LogFail("Handlers must be created in a project")
```
cohix, line 42: Let's make this 'handlers can only be created in projects with a Directive.yaml file'

subo/command/create_handler.go
```diff
@@ -0,0 +1,73 @@
...
+				return errors.Wrap(err, "failed to WriteDirectiveFile")
+			}
+
+			util.LogDone(fmt.Sprintf("handler with resource name %s created", name))
+
+			return nil
+		},
+	}
+
+	cmd.Flags().String(typeFlag, "request", "the method for which you want ")
```
cohix, line 68: Looks like the descriptions for these flags need to be updated

cohix (commented):
Overall the structure looks good! Just a few notes before you move on :)

subo/command/create_handler.go
```diff
@@ -0,0 +1,71 @@
...
+
+			util.LogStart(fmt.Sprintf("creating handler with function name %s", resource))
+
+			bctx, err := context.ForDirectory(dir)
+			if err != nil {
+				return errors.Wrap(err, "failed to ForDirectory")
+			}
+
+			if bctx.Directive == nil {
+				util.LogFail("Handlers can only be created in projects with a Directive.yaml file")
```
cohix, line 41: Since the returned error gets printed as well, I think you could combine this into one line: `return errors.New("cannot create handler, Directive.yaml not found`

subo/command/create_handler.go
```diff
@@ -0,0 +1,71 @@
...
+				return errors.New("Directive.yaml not found")
+			}
```
cohix, line 43: Newline after blocks like this are good to make the code more readable

subo/command/create_handler.go
```diff
@@ -0,0 +1,71 @@
...
+					Type:     handlerType,
+					Resource: resource,
+					Method:   method,
+				},
+			}
+
+			//Add the handler object to the directive file
+			bctx.Directive.Handlers = append(bctx.Directive.Handlers, handler)
+
+			//Write Directive File which overwrites the entire file
```
cohix, line 56: Before writing the file, I would call `bctx.Directive.Validate()` and fail if there are any errors

subo/command/create_handler.go
```diff
@@ -0,0 +1,71 @@
...
+		Short: "create a new handler",
+		Long:  `create a new handler for Subo CLI`,
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			resource := args[0]
+
+			handlerType, _ := cmd.Flags().GetString(typeFlag)
+			method, _ := cmd.Flags().GetString(methodFlag)
+
+			dir, _ := cmd.Flags().GetString(dirFlag)
```
cohix, line 31: I don't think this flag is defined for this command

subo/command/create_handler.go
```diff
@@ -0,0 +1,71 @@
...
+
+			util.LogStart(fmt.Sprintf("creating handler with function name %s", resource))
```
cohix, line 33: I'd say this log should be `"creating handler for %s", resource`

subo/command/create_handler.go
```diff
@@ -0,0 +1,71 @@
...
+
+			//Add the handler object to the directive file
+			bctx.Directive.Handlers = append(bctx.Directive.Handlers, handler)
+
+			//Write Directive File which overwrites the entire file
+			if err := context.WriteDirectiveFile(bctx.Cwd, bctx.Directive); err != nil {
+				return errors.Wrap(err, "failed to WriteDirectiveFile")
+			}
+
+			util.LogDone(fmt.Sprintf("handler with resource name %s created", resource))
```
cohix, line 61: I would make this `"handler for %s created", resource`

subo/command/create_handler.go
```diff
@@ -0,0 +1,71 @@
...
+
+			return nil
+		},
+	}
+
+	cmd.Flags().String(typeFlag, "request", "the kind of input put into the handler")
```
cohix, line 67: I would make the help text `the handler's input type`

subo/command/create_handler.go
```diff
@@ -0,0 +1,71 @@
...
+	cmd.Flags().String(methodFlag, "GET", "used to determine what action to execute on a specific resource")
```
cohix, line 68: This should be `the HTTP method for 'request' handlers`

subo/command/flags.go
```diff
@@ -5,8 +5,12 @@ const (
 	dirFlag             = "dir"
 	namespaceFlag       = "namespace"
 	branchFlag          = "branch"
+	methodFlag          = "method"
+	streamFlag          = "stream"
```
cohix, line 9: This doesn't seem to be used anywhere

subo/command/create_handler.go
MacklinEngineering: We decided against this and will think about implementing on second interation: with steps[].

subo/command/create_handler.go
MacklinEngineering: new line below
Screen Shot 2022-01-28 at 9 43 03 AM

subo/command/create_handler.go
```diff
@@ -0,0 +1,75 @@
...
+	HandlerType string
+	Request     string
+	Method      string
+}
+
+func CreateHandlerCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "handler <resource>",
+		Short: "create a new handler",
+		Long:  `create a new handler for Subo CLI`,
```
cohix, line 24: this should be 'create a new handler in Directive.yaml` or something

cohix (approved):
One last tweak otherwise lgtm!

PR #183, opened 2022-02-01
[Issue-182] Adds a workflow to dispatch data on release

### LINKED ISSUE #182: Subo says there's an update available, brew does not have the updated version
I installed subo a few minutes ago using [homebrew](https://github.com/suborbital/subo#macos-homebrew). Running the command without any arguments told me that there's a new version of subo available:

```shell
$ ℹ️  An upgrade for subo is available: 0.3.0 → 0.3.1
```

Running `brew update && brew upgrade` however did nothing with relation to subo, which makes me think that brew does not know about version 0.3.1 of subo yet.

I'm unsure whether this is a pipeline issue, or a brew issue at this moment.

Related:
* https://github.com/suborbital/subo/issues/139
* https://github.com/suborbital/subo/issues/106

javorszky:
Closes #182 

Adds workflow on release for subo to send data to the homebrew-subo repository to update the homebrew formula.

The three pieces of data to be sent are tag, sha256 of the release, and the url of the newly created release.

These are used in the formula. See https://github.com/suborbital/homebrew-subo/pull/4

The only niggle is that the job needs a personal access token with the public_repo scope. See https://github.com/peter-evans/repository-dispatch#token

That needs to be saved as REPOTOKEN.

### DIFF
--- a/.github/workflows/dispatch-on-release.yml
+++ b/.github/workflows/dispatch-on-release.yml
@@ -1,0 +1,27 @@
+name: dispatch-on-release
+on:
+  release:
+    types: [released]
+
+jobs:
+  post-on-release:
+    name: Post dispatch on release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Stores new tag in environment variable
+        run: |
+          version=`echo ${{ github.event.release.tag_name }} | sed s/v//`
+          echo "version=$version" >> $GITHUB_ENV
+      - name: Downloads release file and computes sha256
+        run: |
+          curl -L -o subo.tar.gz https://github.com/suborbital/subo/archive/v${{ env.version }}.tar.gz
+          suboSHA=`sha256sum subo.tar.gz | cut -c -64`
+          rm subo.tar.gz
+          echo "sha=$suboSHA" >> $GITHUB_ENV
+      - name: Repository Dispatch
+        uses: peter-evans/repository-dispatch@v1
+        with:
+          token: ${{ secrets.REPOTOKEN }}
+          repository: suborbital/homebrew-subo
+          event-type: subo-release
+          client-payload: '{"tag": "${{ env.version }}", "sha": "${{ env.sha }}", "release_url": "${{ github.event.release.html_url}}"}'


jagger27 (approved):

cohix:
I thought the `github_token` that every action runner has access to has the `public_repo` scope?

javorszky:
> I thought the github_token that every action runner has access to has the public_repo scope?

Sadly it isn't. I [updated the test workflow](https://github.com/javorszky/workflow-test-source/commit/76da4118379c1c9ecb7b5650b47cd2a6d697c03d) to use the github token, and got a r issue in the source inaccessible to the integration error, something that did not happen in the previous successful ones: https://github.com/javorszky/workflow-test-source/runs/5033773891?check_suite_focus=true

Github says there are a few ways to reach the token here: https://docs.github.com/en/actions/security-guides/automatic-token-authentication

I tried both `secrets.GITHUB_TOKEN` and `github.token`, and both had the inaccessible issue.

The default permissions for the github token available to the workflows are these: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

We can also elevate / lower the permissions of the token from within the workflow itself: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token

The [permissions needed for the dispatch endpoint](https://docs.github.com/en/rest/reference/repos#create-a-repository-dispatch-event) are either
* personal access token with repo scope, or
* github app token with both metadata:read and contents:read&write

I tried with [setting the contents permission to write](https://github.com/javorszky/workflow-test-source/commit/aaa36b3460034bb93a1d35bd9191d01c70a5864a) as well (metadata is always at least read), but [got the same inaccessible error](https://github.com/javorszky/workflow-test-source/runs/5033898024?check_suite_focus=true).

I think the reason for that is when I elevate the contents to write, it still refers to the repository that the workflow is running in, and I'd need to change the context of another repository.

cohix:
@javorszky interesting... I'm ok with creating a personal access token for this, the only issue is that they expire, and need to be replaced regularly, which is something we may forget or neglect.

That said, I created a token (set to expire in 6 months), and added it to this repo under `RELEASE_GITHUB_TOKEN`... go nuts :)

cohix:
Updated the name of the secret to be `REPOTOKEN`

Please add something to the `knowledge base` in Notion about what this is, what it's used for, and how to refresh it :)

javorszky:
Awesome, thank you! I'll get to it soon.

javorszky:
Knowledge base updated. I'm merging this :)

PR #187, opened 2022-02-03
feat(builder): Introduce templating for prereq commands

### LINKED ISSUE #123: Insert template strings into prereq strings
This would be useful for languages like Grain that reference versions in the prereq steps.

javorszky:
Closes #123 

Adds templating to prerequisite commands

* moves the logic to do that onto the `Prereq` struct and out of the builder
* adds tests
* adds make target for easier local vendoring (not related to functionality)
* reorders imports

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -259,13 +259,25 @@ func (b *Builder) checkAndRunPreReqs(runnable context.RunnableDir, result *Build
 	}
 
 	for _, p := range preReqs {
+
 		filepath := filepath.Join(runnable.Fullpath, p.File)
 
 		if _, err := os.Stat(filepath); err != nil {
 			if errors.Is(err, os.ErrNotExist) {
 				b.log.LogStart(fmt.Sprintf("missing %s, fixing...", p.File))
 
-				outputLog, err := util.RunInDir(p.Command, runnable.Fullpath)
+				cmdTmpl, err := template.New("cmd").Parse(p.Command)
+				if err != nil {
+					return errors.Wrapf(err, "failed to parse prerequisite Command string into template: %s", p.Command)
+				}
+
+				var fullCmd strings.Builder
+				err = cmdTmpl.Execute(&fullCmd, runnable)
+				if err != nil {
+					return errors.Wrap(err, "failed to execute prerequisite Command string with runnableDir")
+				}
+
+				outputLog, err := util.RunInDir(fullCmd.String(), runnable.Fullpath)
 
 				result.OutputLog += outputLog + "\n"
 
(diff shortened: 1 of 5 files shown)


builder/builder.go
```diff
@@ -259,13 +260,19 @@ func (b *Builder) checkAndRunPreReqs(runnable context.RunnableDir, result *Build
...
 		filepath := filepath.Join(runnable.Fullpath, p.File)
 
 		if _, err := os.Stat(filepath); err != nil {
 			if errors.Is(err, os.ErrNotExist) {
 				b.log.LogStart(fmt.Sprintf("missing %s, fixing...", p.File))
 
-				outputLog, err := util.RunInDir(p.Command, runnable.Fullpath)
+				fullCmd, err := p.GetCommand(runnable)
+				if err != nil {
+					return errors.Wrap(err, "failed to get templated full command")
```
ospencer, line 272: It maybe seems a little silly, but the convention we use for error messages is `failed to` + the name of the function that failed. Something like this:
```suggestion
					return errors.Wrap(err, "failed to GetCommand for preReq")
```

ospencer (approved):
This looks great, and tests look handy too! As for the PR title, the scopes typically relate to the subparts of that particular project, usually the folder it's in. In this case, it'd be `feat(builder)`. We don't yet have any of the automation set up that ties any of it together, but that'll be coming soon!

builder/builder.go
javorszky: I'll change that up. Also somewhat relatedly, what do you think of using this? https://thanos.io/tip/contributing/coding-style-guide.md/#wrap-errors-for-more-context-dont-repeat-failed--there

> NOTE: never prefix wrap messages with wording like failed ... or error occurred while.... Just describe what we wanted to do when the failure occurred. Those prefixes are just noise. We are wrapping error, so it’s obvious that some error occurred, right? (: Improve readability and consider avoiding those.

javorszky: I changed up the errors in [64cb16d](https://github.com/suborbital/subo/pull/187/commits/64cb16de97cd55150339a1e381794ed04c531a91).

Additionally I see that `result.OutputLog` gets the log from `util.RunInDir` but it's not being used anywhere that I can see. A global search doesn't reveal anything where `OutputLog` gets passed to a log function call, or even to an fmt.

javorszky:
Hm, interesting that a new commit does not dismiss the approval. 🤔

builder/builder.go
ospencer: @javorszky I'm a fan of that honestly. Will ping @cohix for an opinion as well since he came up with the format that's used currently.

ospencer (approved):

PR #189, opened 2022-02-03
[Issue 188] Add golangci-lint to subo

### LINKED ISSUE #188: (chore) Introduce golangci-lint rules
To make the codebase adhere to idiomatic go style and have consistency between suborbital projects and to guard us against bugs that are easy to miss, we should introduce a set of golangci-lint rules, add a github workflow to check against that, and slowly reformat the codebase to adhere to it

javorszky:
Closes #188


javorszky:
So this one fails because of deadcode issues in the following three places:

* [this handlerData](https://github.com/suborbital/subo/blob/main/subo/command/create_handler.go#L14) is not being used anywhere, and it's not exported. Normally I would remove it, but seeing that it _looks_ important, I want to hold off on it
* [this resourceFlag](https://github.com/suborbital/subo/blob/main/subo/command/flags.go#L12) is not being referenced anywhere either
* nor is [this updateFlag](https://github.com/suborbital/subo/blob/main/subo/command/flags.go#L19)

Can y'all provide more context on what they are (supposed to be) used for? 🙂

javorszky:
Okay, so after a bit more digging

## handlerData

There is no mention of anywhere else in the codebase. It is also not exported, so nothing that imports subo could potentially use it anyways.

I recommend removing it.

## resourceFlag
This is not exported from the `subo/command` package, so scn can't use it. The usage of the property is not found in the entirety of my projects folder which has most of the suborbital repos cloned locally. This is truly not being used anywhere.

I recommend removing it.

## updateFlag
This is being used in two other places within subo itself: [here](https://github.com/suborbital/subo/blob/main/subo/command/compute_deploy_core.go#L41) and [here](https://github.com/suborbital/subo/blob/main/subo/command/compute_deploy_core.go#L185), but both places are commented out.

They were removed in this commit: https://github.com/suborbital/subo/commit/198ad21b40fa56a7fcf90932e215a56b30e77b4f

This is not being used in any of the other projects either. Recommend removing and maybe adding a @todo comment?

javorszky:
I removed the unused pieces of code and created issues to track them:
* updateFlag: https://github.com/suborbital/subo/issues/191
* resourceFlag: https://github.com/suborbital/subo/issues/192
* handlerData: https://github.com/suborbital/subo/issues/193

The linter `unparam` also flagged up a bunch of issues in the docs builder. The function `getCodeText` received both the `docTemplate` and `actionTemplate`, but every caller passed in a constant which doesn't change.

Commit https://github.com/suborbital/subo/pull/189/commits/d6645742a4da35539e61c453bc8941d7ef20699f moves them out from being function arguments so the constants themselves can be used as args in whatever functionality needs them within the `getCodeText` function.

An alternative solution to the unparam, if we want to keep them as function arguments, is to write a test to the `getCodeText` function and pass in different templates for both and test that the text that comes out is what we expect to happen given the template we passed in. cc @denopink

denopink:
@javorszky that totally works for me. We do need the params s.t. it could be used for other types of docs (requiring us to swap out the template/action).
When you say tests, are we thinking go tests? atm subo's tests are written and ran in GitHub (more like integration tests):
https://github.com/suborbital/subo/blob/9249bb720ea332e017190606a247d5cf25b5c4ca/.github/workflows/sanity.yml#L29-L37

I can totally take care of the integration test for `subo doc`, but should we start introducing unit tests starting with `getCodeText` (to resolve the unparam)?
I'm all for it and if we're looking to write unit tests for the other portions of subo, I would love to help with that as well.

javorszky:
@denopink Ideally we should be unit testing every piece of code as well. I'm a fan of table driven tests, and the stretchify/assert package (this one: https://github.com/stretchr/testify), and mockery (https://github.com/vektra/mockery) if we need to create controllable mocks of interfaces. Let me know if I can help with any of it.

I tried creating a test function for the `getCodeText` function in the same directory in a file called `docs_test.go`, but as I'm not entirely familiar with all the pieces that the functions accept and what data looks like, I opted for the easier removal of arguments while still preserving the functionality. If you'd like I can push up a stub test file.

javorszky:
Also, thank you for the link to the integration tests 🙂

javorszky:
@denopink awesome. I've opened https://github.com/suborbital/subo/issues/194 to track readding them with tests.

denopink:
I haven't done table driven tests outside of basic tutorials, so I would love to help with that.
that includes interface mockeries as well!

javorszky:
All right, I'll put some documentation / examples together and share it here later today :)

Makefile
```diff
@@ -20,5 +20,7 @@ mod/replace/atmo:
 
 tidy:
 	go mod tidy && go mod download && go mod vendor
+lint:
```
jagger27, line 23: Let's add `lint` to the prerequisites of `subo` and all of the `subo/*` targets. I don't think it will get in the way of any CI steps.

javorszky: good idea, thank you! fixed in https://github.com/suborbital/subo/pull/189/commits/72e7d42b68f25969a4cc64c0f0650033bbf411fa

PR #195, opened 2022-02-07
feat: Add TS template

### LINKED ISSUE #179: Split JS and TS languages
We'll want to have two separate languages that Subo will know how to scaffold and build, JS and TS. They will share the same builder image, but use different templates. The language pairs will be `javascript`/`js` and `typescript`/`ts`. This will be a breaking change.

ospencer:
Closes #179


templates/typescript/src/index.ts
```diff
@@ -0,0 +1,22 @@
...
+  var TextEncoder: any;
+  var TextDecoder: any;
+}
+
+const decoder = new TextDecoder();
+
+export function run_e(payload: ArrayBuffer, ident: number) {
+  // Imports will be injected by the runtime
+  // @ts-ignore
+  setup(this.imports, ident);
```
cohix, line 16: Why do the imports need to be passed every time run_e is called?

cohix (commented):
looks good, just one question

templates/typescript/src/index.ts
ospencer: That was done just to keep the library setup simple (instead of having separate setup functions for the imports and the ident). It's just doing one assignment, so no real performance hit. We can change that in a future version though.

cohix: Ok, please file an issue for that change

cohix (approved):

templates/typescript/src/index.ts
ospencer: https://github.com/suborbital/reactr/issues/207

PR #196, opened 2022-02-07
feat: Add JS template

ospencer:


templates/javascript/src/index.js
```diff
@@ -0,0 +1,16 @@
+import "fastestsmallesttextencoderdecoder-encodeinto/EncoderDecoderTogether.min.js";
+import { run } from "./lib";
+
+import { setup, runnable } from "@suborbital/runnable";
+
+const decoder = new TextDecoder();
+
+export function run_e(payload, ident) {
+  // Imports will be injected by the runtime
+  setup(this.imports, ident);
```
cohix, line 10: Same question as TS, why do the imports need to be passed every time run_e is called?

templates/javascript/tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
```
cohix, line 1: Why does the JS runnable need a tsconfig?

templates/javascript/src/index.js
ospencer: Answered on the TS PR.

cohix (approved):

Issue #53, opened 2021-03-31
Check Directive for route conflicts at build time


cohix:
When building an app, Subo should attempt to build the Vektor router from the Directive's routes and if Vektor panics, Subo should return a build error.

PR #208, opened 2022-02-19
Validate bundle directive (#53)

rnpridgeon:
Compliments: https://github.com/suborbital/atmo/pull/137

Subo does not currently verify the directive when bundling projects which may result in unexpected errors when deploying a bundle.

This PR adds an additional step to the build process which will validate the augmented directive. 

User defined directive
```
handlers:
  - type: request
    resource: /hello
    method: POST
    steps:
      - fn: helloworld
  - type: request
    resource: /hello
    method: POST
    steps:
      - fn: helloworld
```

Without this change:
```
subo build .
ℹ️  🐳 using Docker toolchain
⏩ START: building runnable: helloworld (rust)
 Downloading crates ...
...
✅ DONE: helloworld was built -> /root/runnable/helloworld/helloworld.wasm
✅ DONE: bundle was created -> /home/ryanp/Workspaces/suborbital/subo/demo/runnables.wasm.zip @ v0.1.0
```

With this change:
```
subo build .
ℹ️  🐳 using Docker toolchain
⏩ START: building runnable: helloworld (rust)
 Downloading crates ...
 ...
✅ DONE: helloworld was built -> /root/runnable/helloworld/helloworld.wasm
Error: failed to Bundle: 🚫 failed to AugmentAndValidateDirectiveFns: found 1 problems:
        a handle is already registered for path '/hello'
Usage:
  subo build [dir] [flags]

```


rnpridgeon:
Turns out this is a bit more aggressive than it should be. Will revisit monday

builder/context/buildcontext.go
```diff
@@ -73,7 +73,9 @@ func ForDirectory(dir string) (*BuildContext, error) {
 
 	directive, err := readDirectiveFile(fullDir)
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to readDirectiveFile")
```
rnpridgeon, line 76: note: 

I moved this out of the `readDirectiveFile` method to the build context because it felt out of place. 

IMO if the caller has expressed the intent to read the directive but it can't be found this is an error; it is the caller's (build context) responsibility to decide what they want to do in this event.

scripts/sanity_project.sh
```diff
@@ -0,0 +1,30 @@
...
+    echo "Error $1 occurred on $2"
+  fi
+
+  # return to origin, clear directory stack
+  pushd -0 > /dev/null && dirs -c
+
+  if [[ -d "$TEST_PROJECT" ]]; then
+    echo "cleaning up test artifacts"
+    # target directory created in docker has root owner
+    sudo rm -rf "$TEST_PROJECT"
```
rnpridgeon, line 20: Note: This is sort of annoying but the docker build context ends up doing everything as root which means the filesystem objects are ultimately owned by root. 

This is worth addressing in another PR.

builder/context/directive.go
```diff
@@ -35,10 +35,6 @@ func readDirectiveFile(cwd string) (*directive.Directive, error) {
 
 	_, err := os.Stat(filePath)
 	if err != nil {
-		if os.IsNotExist(err) {
```
rnpridgeon, line 38: note:

I moved this out of the readDirectiveFile method to the build context because it felt out of place.

IMO if the caller has expressed the intent to read the directive but it can't be found this is an error; it is the caller's (build context) responsibility to decide what they want to do in this event.

scripts/sanity_project.sh
jagger27: I'm not seeing this issue on macOS (probably due to how Docker volumes just straight up ignore permissions on macOS).

javorszky:
The very first piece of review of whether we should find a different expression for "sanity". Reason: https://www.selfdefined.app/definitions/sanity-check/ (tldr it's ableist)

scripts/sanity_project.sh
rnpridgeon: fair enough, I'll back it out and leave a comment about how this may run into permission errors on cleanup

scripts/sanity_runnable.sh
```diff
@@ -0,0 +1,34 @@
...
+
+trap 'catch $? $LINENO' EXIT
+
+catch() {
+  if [[ "$1" != "0" ]]; then
+    echo "Error $1 occurred on $2"
+  fi
+
+  # return to origin, clear directory stack
+  pushd -0 > /dev/null && dirs -c
```
javorszky: TIL about `pushd` and `popd`! https://opensource.com/article/19/8/navigating-bash-shell-pushd-popd

Thank you! 🙂

scripts/sanity_runnable.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# enter project directory
+pushd "$TEST_PROJECT" > /dev/null
+
+# create a runnable for each supported language
+subo create runnable rs-test --lang rust
+subo create runnable swift-test --lang swift
+subo create runnable as-test --lang assemblyscript
+subo create runnable tinygo-test --lang tinygo
+subo create runnable js-test --lang javascript
```
javorszky: if my understanding is correct, if this completes successfully, the files would be left in place, rather than cleaned up after.

Is that expected / desired?

scripts/sanity_runnable.sh
rnpridgeon: one of the most useful bash builtins

scripts/sanity_runnable.sh
rnpridgeon: The trap is triggered on exit (error or otherwise) which cleans up the root project directory if present.

set -e tells the script to terminate on error, set -u tells the script to error if a variable isn't bound to a value (useful when you have potentially dangerous commands like `rm -rf $var`

Makefile
```diff
@@ -15,6 +15,14 @@ subo/docker:
 subo/docker/publish:
 	docker buildx build . --platform linux/amd64,linux/arm64 -t suborbital/subo:dev --push
 
+subo/sanity: subo/sanity/project subo/sanity/runnable
```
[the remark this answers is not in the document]
rnpridgeon: yep,  I had not considered this before but it makes sense. I'm not sure if we can change `sanity.yml` from which I borrowed this term but I have update the scripts and target names accordingly.

rnpridgeon:
renamed targets and scripts to `smoketest`, removed sudo from clean up trap. 

I'm not sure if `sanity.yml` is a required name for the CI so I'll leave that as is for now and we can follow up on it later. 

https://github.com/suborbital/subo/issues/212

scripts/sanity_runnable.sh
rnpridgeon: I do need to comment out javascript for  now since it will fail to bundle until the new images are released

javorszky: Had a conversation on discord, the `trap` takes care of running the `catch` function to clean up whatever happens (error or success)

javorszky (approved):
Thanks for fixing these! 🙂 

:shipit:

PR #213, opened 2022-02-23
fix(ci): Replace deprecated linter step

jagger27:
As per the `golangci-lint` docs, `local-prefixes` is marked as deprecated.

See: https://golangci-lint.run/usage/linters/#gci

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -18,6 +18,10 @@ linters:
 
 linters-settings:
   gci:
-    local-prefixes: github.com/suborbital
+    sections:
+      - standard
+      - default
+      - prefix(github.com/suborbital)
+
   godot:
     scope: all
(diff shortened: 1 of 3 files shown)


javorszky:
@jagger27 it's not deprecated in the version of golangci-lint that the workflow is using: https://github.com/suborbital/subo/blob/main/.github/workflows/golangci-lint.yml#L22

1.44.2, yes

To fix the brokenness, upgrade the version of golangci-lint as well :)

.golangci.yml
```diff
@@ -18,6 +18,10 @@ linters:
 
 linters-settings:
   gci:
-    local-prefixes: github.com/suborbital
+    sections:
+      - standard
+      - default
+      - prefix(github.com/suborbital)
```
javorszky, line 24: now comes the better part, do we want to separate this to further?

```
- prefix(github.com/suborbital)
- prefix(github.com/suborbital/subo)
```

javorszky (approved):
Minor question, but approved otherwise

.golangci.yml
jagger27: We'll bike shed this tomorrow! I think I'm okay with three sections.

PR #217, opened 2022-02-25
fix(builder): Only run Javy once for JS/TS Runnables

### LINKED ISSUE #216: JS builder: Javy always run twice
The build script runs Javy and so does subo—only one needs to!

ospencer:
Closes #216

### DIFF
--- a/builder/context/native.go
+++ b/builder/context/native.go
@@ -40,11 +40,9 @@ var nativeCommandsForLang = map[string]map[string][]string{
 		},
 		"typescript": {
 			"npm run build",
-			"javy build/index.js -o {{ .Name }}.wasm",
 		},
 		"javascript": {
 			"npm run build",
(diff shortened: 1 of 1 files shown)


cohix (approved):

builder/context/native.go
```diff
@@ -40,11 +40,9 @@ var nativeCommandsForLang = map[string]map[string][]string{
 		},
 		"typescript": {
 			"npm run build",
```
cohix, line 42: Hmm, the `webpack` step is in a `prebuild` script... why does that run even though it's not being called?

ospencer: Heh, I probably shouldn't bore you with all of the details of the lifecycles of npm scripts, but `pre` scripts get run before the script that shares a name with it; `prebuild` -> `build`. It's actually pretty handy for cases like this, where subo now only needs to worry about running the `build` command.

PR #219, opened 2022-03-01
Missing error assignment leads to missed error checking #185

### LINKED ISSUE #185: Missing error assignment leads to missed error checking
Here: https://github.com/suborbital/subo/blob/main/builder/template/config.go#L38

```go
	config, err := os.UserConfigDir()
	if err != nil {
		return "", errors.Wrap(err, "failed to get UserConfigDir")
	}

	tmplPath := filepath.Join(config, "suborbital", "templates")

	if os.Stat(tmplPath); err != nil {
		//
	}
```
The `err != nil` check after os.Stat is using the `err` variable from `or.UserConfigDir`, and we already guaranteed that's always going to be nil by the time execution gets to `os.Stat`.

Should be
```
if err := os.Stat(tmplPath); err != nil {
```

yashikajotwani12:
Fixes #185

### DIFF
--- a/builder/template/config.go
+++ b/builder/template/config.go
@@ -36,7 +36,7 @@ func TemplateRootDir() (string, error) {
 
 	tmplPath := filepath.Join(config, "suborbital", "templates")
 
-	if _, err := os.Stat(tmplPath); err != nil {
+	if _, err = os.Stat(tmplPath); err != nil {
 		if errors.Is(err, os.ErrNotExist) {
 			if err := os.MkdirAll(tmplPath, util.PermDirectory); err != nil {
 				return "", errors.Wrap(err, "failed to MkdirAll template directory")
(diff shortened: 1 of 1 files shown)


jagger27 (approved):

builder/template/config.go
```diff
@@ -36,7 +36,7 @@ func TemplateRootDir() (string, error) {
 
 	tmplPath := filepath.Join(config, "suborbital", "templates")
 
-	if os.Stat(tmplPath); err != nil {
+	if _, err := os.Stat(tmplPath); err != nil {
```
javorszky, line 39: Another very tiny fix here is that rather than instantiate and assign the err variable (`:=`), we can skip the instantiate part and only reassign, because a variable with the name `err` (and the same type) already exists.

So instead of `if _, err := os.Stat...`, we can do `if _, err = os.Stat...`. Note the missing `:`. That way we're not going to shadow an earlier declared variable :)

javorszky (changes requested):

javorszky (approved):

PR #222, opened 2022-03-11
Simplify CI and Releases

arbourd:
Condenses the CI and Release workflows into a single workflow called "CI" that:
- golang-lints for each PR, commit to `main` and tags
- runs tests for each PR, commit to `main` and tags
- releases on tag creation with GoReleaser

Also adds a `brews` stanza to the GoReleaser configuration that updates the Homebrew definition for subo, automatically.


.goreleaser.yml
```diff
@@ -1,16 +1,14 @@
...
+  - goos:
       - linux
       - darwin
     goarch:
       - amd64
       - arm64
+    env:
+      - CGO_ENABLED=0
+    ldflags:
+      - -s -w -X main.VERSION={{ .Version }}
```
arbourd, line 11: Potential for adding other flags here if needed.

.github/workflows/ci.yml
```diff
@@ -48,3 +55,31 @@ jobs:
...
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
+        with:
+          go-version: 1.17
+
+      - uses: tibdex/github-app-token@v1
+        id: generate_token
+        with:
+          app_id: ${{ secrets.APP_ID }}
+          private_key: ${{ secrets.APP_PRIVATE_KEY }}
```
arbourd, line 76: Uses a GitHub application to generate single use GH tokens that can be used to do work across repos (as the GH_TOKEN cannot by default)

.github/workflows/ci.yml
```diff
@@ -48,3 +55,31 @@ jobs:
       - name: Check TinyGo version number consistency
         run: |
           builder/docker/tinygo/sanity.sh
+
+  # only run if reference is a tag
+  release:
+    if: startsWith(github.ref, 'refs/tags/v')
+
+    needs: [lint, test]
+    runs-on: ubuntu-latest
```
arbourd, line 64: Will only run this step if the target/reference is a tag (which implies release or pre-release). This step will also only run if both `lint` and `test` steps have passed.

cohix (commented):
Woah fancy. I didn't know goreleaser could do brew taps. I'm going to let @javorszky approve this one since he did the original brew deploy workflow.

.goreleaser.yml
```diff
@@ -20,3 +18,23 @@ checksum:
...
+    folder: Formula
+
+    tap:
+      owner: suborbital
+      name: homebrew-subo
+      token: '{{ .Env.HOMEBREW_TAP_GITHUB_TOKEN }}'
+
+    commit_author:
+      name: suborbital
+      email: 12345678+suborbital[bot]@users.noreply.github.com
```
arbourd, line 35: Need to confirm this number before merge.

.goreleaser.yml
javorszky: We usually had version, sha, and date compiled passed onto it and expose all of that in the /status endpoint, or even the /health endpoint

.goreleaser.yml
javorszky: Where does this number come from?

.github/workflows/ci.yml
```diff
@@ -48,3 +55,31 @@ jobs:
       - name: Check TinyGo version number consistency
         run: |
           builder/docker/tinygo/sanity.sh
+
+  # only run if reference is a tag
+  release:
+    if: startsWith(github.ref, 'refs/tags/v')
```
javorszky, line 61: given that the workflow only runs on PRs against main with tags that begin with `v` as described here
```
on:
  pull_request:
  push:
    branches:
      - main
    tags:
      - v*
```

Do we need this?

javorszky (changes requested):
Oh this is super exciting! I love that it does the brew release, and calculates the necessary hashes too! At least if I'm reading https://goreleaser.com/customization/homebrew/ right.

There is one question about the "run if" part, but otherwise looks good!

ospencer (commented):
Very nice!

.github/workflows/ci.yml
arbourd: We do. At the top level, it means "run on every pull request, commit on main and tag" whereas, the if block only runs the `release` job only if it is a tag, but will wait for tests to pass based on the `need` block.

.goreleaser.yml
arbourd: I believe its randomly generated and I'll need to get the bot to commit something first to find out what it is.

javorszky (approved):

.goreleaser.yml
arbourd: ya let me add that, thats the default for GoReleaser too :)

javorszky:
Oh, one more question. Will it push to the main branch of the homebrew repo, or open a PR?

arbourd:
@javorszky It will just push: https://github.com/goreleaser/goreleaser/issues/2870

.goreleaser.yml
arbourd: It stays but it can break. It's kind of a weird hack to use a bot as the committer. We could use a static account too. Do you have any other ideas for who should be doing the writing?

javorszky: we could create an automation@suborbital.dev account? hm... I'm not sure actually

arbourd: I actually changed this to be the original committers name and email. That means that the author of the tag should get the credit on that PR. They will also get a push notification if it were to fail in CI. I think this is best.

What do you think?

javorszky: Nice! Didn't know we could do that :)

PR #229, opened 2022-03-20
add Cloudflare deployer

xtuc:


deployer/cloudflaredeployer.go
```diff
@@ -0,0 +1,281 @@
...
+	"github.com/suborbital/subo/project"
+	"github.com/suborbital/subo/subo/util"
+)
+
+const (
+	cloudflareDeployJobType = "cloudflare"
+	routerScriptName        = "subo-router"
+)
+
+//go:embed cloudflare/templates/request_worker.js
```
cohix, line 27: All of these template files should be put into the `templates` directory and the template renderer should be used to write them into the `.deploy` directory when they need to be used (which is a Subo norm)

project/context.go
```diff
@@ -38,6 +41,7 @@ type Context struct {
 	MountPath     string
 	RelDockerPath string
 	BuilderTag    string
+	AppSource     appsource.AppSource
```
cohix, line 44: Instead of adding AppSource to Context, we should create a local instance of it where it's needed instead

go.mod
```diff
@@ -44,5 +45,8 @@ require (
 	golang.org/x/crypto v0.0.0-20220307211146-efcb8507fb70 // indirect
 	golang.org/x/net v0.0.0-20220225172249-27dd8689420f // indirect
 	golang.org/x/text v0.3.7 // indirect
+	golang.org/x/time v0.0.0-20220224211638-0e9765cccd65 // indirect
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
 )
+
+replace github.com/cloudflare/cloudflare-go => ./deps/cloudflare-go
```
cohix, line 52: This should be replaced by a GitHub fork instead of something local (or preferably changes get upstreamed)

deployer/cloudflaredeployer.go
```diff
@@ -0,0 +1,281 @@
...
+	wasmReader := bytes.NewReader(data)
+
+	scriptParams := cloudflare.WorkerScriptParams{
+		Script: script,
+		Bindings: map[string]cloudflare.WorkerBinding{
+			"WASM_MODULE": cloudflare.WorkerWebAssemblyBinding{Module: wasmReader},
+		},
+	}
+
+	scriptName := fmt.Sprintf("subo-%s-%s", _type, runnable.Name)
```
cohix, line 227: This should probably be based on the `ApplicationIdentifier` instead of `subo-` so that it's unique?

deployer/cloudflaredeployer.go
xtuc: done

project/context.go
xtuc: done

deployer/cloudflaredeployer.go
xtuc: done

PR #232, opened 2022-03-29
add Cloudflare deployer

xtuc:
Originally reviewed in https://github.com/suborbital/subo/pull/229

### DIFF
--- a/deployer/cloudflaredeployer.go
+++ b/deployer/cloudflaredeployer.go
@@ -1,0 +1,312 @@
+package deployer
+
+import (
+	"bytes"
+	"context"
+	_ "embed"
+	"fmt"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+	"text/template"
+
+	cloudflare "github.com/cloudflare/cloudflare-go"
+	"github.com/pkg/errors"
+
+	"github.com/suborbital/atmo/atmo/appsource"
+	"github.com/suborbital/atmo/atmo/options"
+	"github.com/suborbital/atmo/directive"
+	subotemplate "github.com/suborbital/subo/builder/template"
+	"github.com/suborbital/subo/project"
+	"github.com/suborbital/subo/subo/util"
+)
+
+const (
+	cloudflareDeployJobType = "cloudflare"
+	routerScriptName        = "subo-router"
+
+	bundlePath = "./runnables.wasm.zip"
+)
+
+//go:embed cloudflare/shared.js
+var shared string
+
+//go:embed cloudflare/itty-router.js
+var ittyRouter string
+
+type CloudflareDeployJob struct {
+	domain          string
+	repo            string
+	branch          string
+	updateTemplates bool
+}
+
+func NewCloudflareDeployer(repo, branch, domain string, updateTemplates bool) DeployJob {
+	c := &CloudflareDeployJob{
+		repo:            repo,
+		branch:          branch,
+		domain:          domain,
+		updateTemplates: updateTemplates,
+	}
+	return c
+}
+
+func (k *CloudflareDeployJob) Type() string {
+	return cloudflareDeployJobType
+}
+
+type cfWorker struct {
+	Path           string
+	Method         string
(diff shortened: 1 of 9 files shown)


jagger27:
Hi @xtuc, sorry for the delay. Could you try pulling in the latest main? The linter is being a bit aggressive here (we missed it because this specific check is disabled in go1.18).

deployer/cloudflaredeployer.go
```diff
@@ -0,0 +1,312 @@
...
+	cloudflareDeployJobType = "cloudflare"
+	routerScriptName        = "subo-router"
+
+	bundlePath = "./runnables.wasm.zip"
+)
+
+//go:embed cloudflare/shared.js
+var shared string
+
+//go:embed cloudflare/itty-router.js
```
cohix, line 35: Is there any reason why these should be embedded instead of included in the `templates/cf` directory?

deployer/cloudflaredeployer.go
```diff
@@ -0,0 +1,312 @@
...
+
+	if job.updateTemplates {
+		templatesPath, err = subotemplate.UpdateTemplates(job.repo, job.branch)
+		if err != nil {
+			return errors.Wrap(err, "🚫 failed to UpdateTemplates")
+		}
+	}
+
+	for _, meta := range appSource.Applications() {
+		log.LogStart(fmt.Sprintf("deploying %s (%s) runnables as workers", meta.Identifier, job.domain))
```
cohix, line 101: I think this could just be `deploying %s (%s) to Cloudflare Workers`

deployer/cloudflaredeployer.go
```diff
@@ -0,0 +1,312 @@
...
+		workersToRoute := make(map[string]cfWorker)
+
+		for _, handler := range appSource.Handlers(meta.Identifier, meta.AppVersion) {
+			if handler.Input.Type != "request" {
+				log.LogWarn(fmt.Sprintf("unsupported handler type: %s; ignoring.", handler.Input.Type))
+				continue
+			}
+
+			if len(handler.Steps) != 1 {
+				log.LogWarn(fmt.Sprintf("only one step is currently supported per handler"))
```
cohix, line 114: I'd like to include information about which handler has more than 1 step, and I think this should be an error that halts the deployment rather than a warning and `continue`

deployer/cloudflaredeployer.go
```diff
@@ -0,0 +1,312 @@
...
+		}
+
+		log.LogDone(fmt.Sprintf("deployed to Cloudflare at %s", routerRoute))
+	}
+
+	return nil
+}
+
+func getAPIClient() (*cloudflare.API, error) {
+	apiToken := os.Getenv("CF_API_TOKEN")
```
cohix, line 159: Could we make this `SUBO_CLOUDFLARE_TOKEN` ? Or is this a standard env key used by other tools?

deployer/cloudflaredeployer.go
```diff
@@ -0,0 +1,312 @@
...
+	accountId := os.Getenv("CF_ACCOUNT_ID")
```
cohix, line 160: Similar, could this be `SUBO_CLOUDFLARE_ACCOUNTID` ?

deployer/cloudflaredeployer.go
```diff
@@ -0,0 +1,312 @@
...
+
+	for _, runnable := range runnables {
+		runnablesByName[runnable.Name] = runnable
+	}
+
+	return runnablesByName
+}
+
+func renderRouterWorker(templatesPath string, workersToRoute map[string]cfWorker) (string, error) {
+	templatePath := path.Join(templatesPath, "cf/router_worker.js.tmpl")
```
cohix, line 186: You can pass multiple parts into `Join`, i.e. `(templatesPath, "cf", "router_worker.js.tmpl")`... this ensures it will work on all platforms.

cohix (commented):
This is looking really good! Just a few things.

PR #233, opened 2022-03-29
Rename sanity.sh 

### LINKED ISSUE #212: If possible rename sanity.yml
It has come to our attention that the term "sanity check" may be insensitive and should be avoided where possible.

https://www.selfdefined.app/definitions/sanity-check/

yashikajotwani12:
Changed to smoke
Fixes #212


jagger27 (approved):
Thank you!

PR #235, opened 2022-04-04
bug: Add missing packages to subo dockerfile

rnpridgeon:
Currently make `subo/docker` fails because it doesn't have access to the complete source. 

This adds the missing packages to the docker file copy directive


MacklinEngineering (approved):
Woohoo! great work.

PR #237, opened 2022-04-06
feat(builder)!: Add configurable JS toolchains

ospencer:


builder/builder.go
```diff
@@ -25,9 +25,20 @@ var dockerImageForLang = map[string]string{
...
 }
 
+// BuildConfig is the configuration for a Builder.
+type BuildConfig struct {
+	JsToolchain string
+}
+
+// DefaultBuildConfig is the default build configuration.
+var DefaultBuildConfig = BuildConfig{
+	JsToolchain: "npm",
```
rnpridgeon, line 35: Should we restrict this to a set of known good values? 

How does the user install alternative tool chains in the docker image?

rnpridgeon: > this configuration can only be set programmatically by using subo as a library

This does mostly make things better though generally its best to fail as early as possible otherwise it might as well be javascript 🔥 

Like its going to be really annoying if I set this stuff up but only build rust for a while then go to build some javascript and this thing that has been working for weeks just barfs. 

It's not a hill I care do die on though so I'll leave it to your discretion

rnpridgeon (approved):
LGEnoughTM

ospencer:
@rnpridgeon Want to take one last look at this? I rebased this on top of the latest changes and want to make sure you think it still looks good.

rnpridgeon (approved):

rnpridgeon (changes requested):
updated review status based on my last comment about adding sanity tests

rnpridgeon (approved):
This is a fair argument granted it would be nice to have a list of known "good" build chains.

Anyway so long as we have the known good supported path tested so we can catch regressions I'm happy

scripts/smoketest.sh
```diff
@@ -31,8 +31,7 @@ subo create runnable rs-test --lang rust
 subo create runnable swift-test --lang swift
 subo create runnable as-test --lang assemblyscript
 subo create runnable tinygo-test --lang tinygo
-# TODO: uncomment when js builder is checked in
-#subo create runnable js-test --lang javascript
+subo create runnable js-test --lang javascript
```
arbourd, line 34: We're not calling this anymore, as a heads up, just changed

arbourd: https://github.com/suborbital/subo/pull/271

PR #238, opened 2022-04-06
chore: Run unit tests in CI

ospencer:


rnpridgeon (approved):

arbourd (approved):

jagger27 (commented):
The deleted files are for testing the `subo docs` command and should be kept for that purpose.

PR #239, opened 2022-04-07
feat: Save environment token if it was provided at the command line

### LINKED ISSUE #141: Auto-save successfully used tokens & allow tokens to be passed in as an env var
Would be great if successfully using a token caused it to be saved locally. Currently, tokens are only saved locally when `subo compute create token` is used.

Also having an option of passing in the token as an env var (maybe expecting it as a base64 encoded secret) for infra-as-code, playbook and debugging purposes would be great.

### LINKED ISSUE #143: Consistent user cache directory
Currently, subo and sat use `os.TempDir()` to store temporary files. I think we should use `os.UserCacheDir()` instead, especially for storing the environment token as that is sensitive. It doesn't particularly matter which one is used, but we should be consistent.

Related: #141, #111 

See also: https://github.com/suborbital/sat/issues/25

jagger27:
Subo now writes tokens passed in on the command line during a fresh run of `subo compute deploy core` to the cache directory. 

This PR also moves the Subo cache directory from `os.TempDir()` to `os.UserCacheDir()`. In the event that the UserCacheDir is unavailable, we can fallback to TempDir. This PR also introduces a migration helper to move existing cached files from the old path to the new one.

Closes #141
Closes #143


javorszky (changes requested):
From the **what does it do** point it looks good.

I think the **when does it do it** could be changed to be either a flag, or a separate command that users could run themselves once if they need to migrate cache.

That way this is isolated, so we can attach a deprecated notice easily to it, and also remove it a few versions down the line by deleting an entire file.

As all of our current software is beta, I think it's okay and acceptable to break BC and invalidate cache, so if users run subo and it looks for cache in a new place and doesn't find it, it's essentially invalidated cache.

subo/util/token.go
```diff
@@ -2,36 +2,44 @@ package util
...
-			return errors.Wrap(err, "failed to Mkdir")
-		}
+	tokenPath, err := getTokenPath()
+	if err != nil {
+		return err
 	}
 
 	if err := ioutil.WriteFile(tokenPath, []byte(tokenStr), PermFilePrivate); err != nil {
-		return errors.Wrap(err, "failed to WriteFile for token")
+		return errors.Wrap(err, "failed to write "+tokenPath)
```
rnpridgeon, line 26:
```suggestion
		return errors.Wrapf(err, "failed to write %s " + tokenPath)
```

rnpridgeon (commented):
I think there are a few things we need to consider with regard to both the existing behavior and proposed behavior. 

1. Should the output directory be configurable ?

The one issue I foresee with sticking it in the user cache is that it's only available to the person who created the environment token. 

This isn't terribly helpful if the environment is shared.

2. Do we want to support the ability to manage multiple environments ?

https://github.com/suborbital/scn/blob/main/compute/network/api/migrations/connor-scn-email-verification.sql

https://github.com/suborbital/scn/blob/main/compute/network/api/migrations/connor-tokens-and-environments.sql

Note the uuid and id are unique though the user and environment need not be. If I understand  correctly this means when a user creates a new environment it does not invalidate the old one which means people may have multiple, active environments. 

The current code seems to overwrite any existing token in the cache (though I may have missed where we check for that).

subo/util/token.go
jagger27: 1. I see this as a feature, not a bug! :^) The previous /var/tmp directory location (on macOS anyway) was also per-user. 

2. I don't think handling multiple environments from `subo`'s perspective is a major concern at this point, since the actual token is copied locally into its own file per `compute deploy` folder (and can be overridden). I agree it's not ideal though.

> The current code seems to overwrite any existing token in the cache (though I may have missed where we check for that).

Yes. If you manually create new a token it will overwrite the old one. The other code path would only be triggered if one wasn't cached.

rnpridgeon (approved):
Seems fine to me though I do think the product related questions are something we need to start defining sooner rather than later.

subo/util/token.go
cohix: Ryan is correct we'll need to handle multiple environments eventually, but it's probably a few months out, and I don't think it'll be a terribly hard change to make. That said, if you want to incorporate that now, I wouldn't be mad.

subo/util/token.go
```diff
@@ -2,36 +2,44 @@ package util
...
-			return errors.Wrap(err, "failed to Mkdir")
-		}
+	tokenPath, err := getTokenPath()
+	if err != nil {
+		return err
 	}
 
 	if err := ioutil.WriteFile(tokenPath, []byte(tokenStr), PermFilePrivate); err != nil {
-		return errors.Wrap(err, "failed to WriteFile for token")
+		return errors.Wrapf(err, "failed to write %s "+tokenPath)
```
javorszky, line 26: this will error because there's a placeholder (`%s`), but there's nothing going in there. That plus needs to be comma :)
```suggestion
		return errors.Wrapf(err, "failed to write %s ", tokenPath)
```

subo/util/token.go
```diff
@@ -2,36 +2,44 @@ package util
...
 
 func WriteEnvironmentToken(tokenStr string) error {
-	tokenPath := getTokenTmpDir()
-	if _, err := os.Stat(tokenPath); os.IsNotExist(err) {
-		if err := os.MkdirAll(filepath.Dir(tokenPath), PermDirectoryPrivate); err != nil {
-			return errors.Wrap(err, "failed to Mkdir")
-		}
+	tokenPath, err := getTokenPath()
+	if err != nil {
+		return err
```
javorszky, line 22: should this be wrapped?

subo/util/token.go
```diff
@@ -2,36 +2,44 @@ package util
...
 	"github.com/pkg/errors"
 )
 
-func getTokenTmpDir() string {
-	tokenPath := filepath.Join(os.TempDir(), "suborbital", "compute", "envtoken")
-	return tokenPath
+func getTokenPath() (string, error) {
+	tokenPath, err := CacheDir("compute")
+	if err != nil {
+		return "", err
```
javorszky, line 13: should this err be wrapped?

subo/util/token.go
```diff
@@ -2,36 +2,44 @@ package util
...
 	}
+
 	return nil
 }
 
 func ReadEnvironmentToken() (string, error) {
-	tokenPath := getTokenTmpDir()
+	tokenPath, err := getTokenPath()
+	if err != nil {
+		return "", err
```
javorszky, line 35: should this err be wrapped?

subo/util/token.go
```diff
@@ -2,36 +2,44 @@ package util
...
+	}
+
 	buf, err := ioutil.ReadFile(tokenPath)
+
 	if err != nil {
-		return "", errors.Wrap(err, "failed to ReadFile for token")
+		return "", errors.Wrap(err, "failed to read "+tokenPath)
```
javorszky, line 41: if the previous wrap is using `Wrapf` with a placeholder, should this be standardised as well?
```suggestion
		return "", errors.Wrapf(err, "failed to read %s", tokenPath)
```

javorszky (changes requested):
Tiny tiny nits, one breaking err to fix, otherwise good to go

javorszky (approved):

PR #243, opened 2022-04-13
fix: Remove update checking when inside Docker

### LINKED ISSUE #242: Subo should not check for updates when running inside of a Docker container
It appears that the root certs in our Docker images are outdated, producing this error:

```
ℹ️  failed to getLatestVersion: failed to fetch latest subo release: Get "https://api.github.com/repos/suborbital/subo/releases/latest": x509: certificate signed by unknown authority
```

Subo doesn't need to check for updates within Docker anyway, so there should be a build flag or runtime flag to disable that feature.

jagger27:
This is done by introducing a go build flag.

(Merge after #239)

Closes #242

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,53 +1,14 @@
 package main
 
 import (
-	"context"
 	"os"
-	"time"
-
-	"github.com/suborbital/subo/subo/release"
-	"github.com/suborbital/subo/subo/util"
 )
 
-const checkVersionTimeout = 500 * time.Millisecond
-
 func main() {
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
-	versionChan := checkVersion(ctx)
-
 	rootCmd := rootCommand()
 	if err := rootCmd.Execute(); err != nil {
 		os.Exit(-1)
 	}
 
-	select {
-	case msg := <-versionChan:
-		if msg != "" {
-			util.LogInfo(msg)
-		}
-	case <-time.After(checkVersionTimeout):
-		util.LogFail("failed to CheckForLatestVersion due to timeout")
-	}
-}
-
-func checkVersion(ctx context.Context) chan string {
-	versionChan := make(chan string)
-
-	go func() {
-		msg := ""
-		if version, err := release.CheckForLatestVersion(); err != nil {
-			msg = err.Error()
-		} else if version != "" {
-			msg = version
-		}
-
-		select {
-		case <-ctx.Done():
-		default:
-			versionChan <- msg
-		}
-	}()
-
-	return versionChan
+	checkForUpdates()
 }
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -1,53 +1,14 @@
...
-const checkVersionTimeout = 500 * time.Millisecond
-
 func main() {
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
-	versionChan := checkVersion(ctx)
-
 	rootCmd := rootCommand()
 	if err := rootCmd.Execute(); err != nil {
 		os.Exit(-1)
```
javorszky, line 10: while here, can you also fix this? The only valid exit codes are 0-255. If this is an error, make it a 1, rather than a negative 1 :)

update_checker.go
```diff
@@ -0,0 +1,44 @@
...
+		}
+
+		select {
+		case <-ctx.Done():
+		default:
+			versionChan <- msg
+		}
+	}()
+
+	util.LogInfo("hhelloo?")
```
javorszky, line 35: Is this a debugging print? 😄

update_checker.go
```diff
@@ -0,0 +1,44 @@
...
+	select {
+	case msg := <-versionChan:
+		if msg != "" {
+			util.LogInfo(msg)
+		}
+	case <-time.After(checkVersionTimeout):
```
javorszky, line 41: I wonder if the `context.WithTimeout(context.Backgroud(), checkVersionTimeout)` could be used here. https://go.dev/play/p/_oG0Kgm7O19

Yes, technically the time.After checkversion timeout does the same thing, but since we already have a context, we might as well use it.

main.go
javorszky: Depending on what the error was with the rootCmd, we could choose between any of the special ones from here: http://www.bic.mni.mcgill.ca/~dale/helppages/BashGuide/advshell/exitcodes.html, or use the catchall 1.

update_checker.go
```diff
@@ -0,0 +1,44 @@
...
+	go func() {
+		msg := ""
+		if version, err := release.CheckForLatestVersion(); err != nil {
+			msg = err.Error()
+		} else if version != "" {
+			msg = version
+		}
+
+		select {
+		case <-ctx.Done():
```
javorszky, line 29: this is confusing here. The only time this is ran is when some time has elapsed and the time.after triggers, the function gets to the end, the deferred cancel is called, and the code already returned from release.checkforlatestversion.

Alternatively what I would do is create a context.withtimeout, pass that one to the checkforlatestversion, propagate it across until github.NewClient.Repositories.GetLatestRelease, because the first argument there is a context.

That way the github sdk will return the "context deadline exceeded" error, and we don't have to deal with that here.

javorszky (changes requested):
The context timeout needs to be looked at, otherwise only tiny nits 😄

update_checker.go
jagger27: Whoops! >_<

javorszky (approved):
Nice! Thank you :)

:shipit:

PR #245, opened 2022-04-15
feat(builder): Vendor rust dependencies for faster builds

rnpridgeon:
without cargo vendor
```
    Finished release [optimized] target(s) in 24.34s
✅ DONE: rusty was built -> /root/runnable/rusty.wasm

real    0m26.111s
user    0m0.090s
sys     0m0.051s

```

with cargo vendor
```
[source.crates-io]
replace-with = "vendored-sources"

[source.vendored-sources]
directory = "vendor"
    Finished release [optimized] target(s) in 0.01s
✅ DONE: rusty was built -> /root/runnable/rusty.wasm

real    0m7.056s
user    0m0.065s
sys     0m0.061s

```

### DIFF
--- a/builder/native.go
+++ b/builder/native.go
@@ -20,7 +20,7 @@ func NativeBuildCommands(lang string) ([]string, error) {
 var nativeCommandsForLang = map[string]map[string][]string{
 	"darwin": {
 		"rust": {
-			"cargo build --target wasm32-wasi --lib --release",
+			"cargo vendor && cargo build --target wasm32-wasi --lib --release",
 			"cp target/wasm32-wasi/release/{{ .UnderscoreName }}.wasm ./{{ .Name }}.wasm",
 		},
 		"swift": {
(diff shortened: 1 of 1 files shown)


arbourd (approved):

builder/native.go
```diff
@@ -20,7 +20,7 @@ func NativeBuildCommands(lang string) ([]string, error) {
 var nativeCommandsForLang = map[string]map[string][]string{
 	"darwin": {
 		"rust": {
-			"cargo build --target wasm32-wasi --lib --release",
+			"cargo vendor && cargo build --target wasm32-wasi --lib --release",
```
cohix, line 23: Let's make these separate entries, they're meant to be executed sequentially, and it'll give more informative errors if things fail.

rnpridgeon: They are still executed sequentially; && is shorthand for "and if" meaning it will execute the subsequent command if and only if the first command completes successfully. 

I do not believe this will result in any less logging as we are not redirecting stdout nor stderr

https://pubs.opengroup.org/onlinepubs/9699919799/

> A ';' separator or a ';' or <newline> terminator shall cause the preceding AND-OR list to be executed sequentially; an '&' separator or terminator shall cause asynchronous execution of the preceding AND-OR list.

PR #263, opened 2022-04-20
feat!: Make build output streamable

ospencer:
This PR includes one small breaking change to the builder's `ForDirectory`—it now accepts an `io.Writer` as the second argument that will have build output written to it. It can be `nil`.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -3,6 +3,7 @@ package builder
 import (
 	"fmt"
 	"html/template"
+	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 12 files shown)


cohix:
This looks great! Could we add a test where you pass in a `bytes.NewBuffer` as the writer, and then run a build to verify it does indeed write into it? Doesn't matter if the build fails.

ospencer:
There are now tests. Tested the command runner directly though instead of running a whole build (didn't want to run docker, etc in the unit tests).

rnpridgeon (approved):

PR #266, opened 2022-04-23
chore(ci): add multi-stage build and action for tinygo-base

arbourd:


builder/docker/tinygo/Dockerfile.base
```diff
@@ -1,17 +1,31 @@
...
-RUN make llvm-source
-RUN git submodule update --init --remote lib/wasi-libc
-RUN export CC=clang-11 && export CXX=clang++-11 && make llvm-build
+RUN git clone --depth 1 --branch v0.22.0 https://github.com/tinygo-org/tinygo.git /root/tinygo && \
+    git clone --depth 1 --branch version_105 https://github.com/WebAssembly/binaryen /root/binaryen && \
+    mkdir /root/binaryen/build && \
+    mkdir /root/runnable && \
+    git submodule update --init && \
+    mv Makefile Makefile.old && \
+    cat Makefile.old | sed "s/\.\/build\/tinygo build-library -target=/.\/build\/release\/tinygo\/bin\/tinygo build-library -target=/" > Makefile
```
arbourd, line 28: This was a suggestion from the tinygo maintainers. It seems that there are 2 release bins available after building: one with C headers statically compiled, another linked to a place where we're not putting them. Changing the Makefile fixes this by using `./build/release/tinygo/bin/tinygo` over `./build/tinygo`

arbourd: Also, grouping these tasks into a single `RUN` ensures they are one layer.

jagger27: I guess this should be okay since we're pinned to a tag!

jagger27 (approved):

.github/workflows/tinygo-base.yml
```diff
@@ -0,0 +1,27 @@
...
+    paths:
+      - .github/workflows/tinygo-base.yml
+      - builder/docker/tinygo/Dockerfile.base
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        platform: [linux/amd64, linux/arm64]
```
arbourd, line 14: The linux/arm64 will always timeout, but I think its worth leaving it here for the time being so that we catch early failures.

This workflow only runs on changes to the tinygo-base Dockerfile, or this action, in a pull-request.

PR #269, opened 2022-04-28
subo create runnable flag for directly exposing on Directive #174

### LINKED ISSUE #174: subo create runnable flag for directly exposing on Directive
Currently `subo create runnable` doesn't touch the Directive at all.

In certain situations it might be useful to append the newly created runnable to the Directive and expose it as a handler straight away. This is a functionality that is being added as #134 as a separate command -- but it might be that it would make sense also as a flag on `create runnable`.

This flag, say `subo create runnable X --handler` would  create the runnable `X` and append it to the directive as a handler `/X`. Alternatively if specified as `... --handler foo`, that would set the handler to `/foo` instead.

The equivalent currently via shell is
```sh
# create the runnable
subo create runnable X
tee -a Directive.yaml << EOF
  - type: request
    resource: /X
    steps:
      - fn: X
EOF
```

yashikajotwani12:
Screenshot from 2022-04-28 20-59-49

Fixes #174


arbourd:
#275 should unblock this

arbourd:
Could you rebase @yashikajotwani12 and let us know if it passes? It might fail due to `GITHUB_TOKEN` -- not sure.

PR #272, opened 2022-04-28
chore(ci): add wat tests

arbourd:


arbourd:
These tests should pass with #279

rnpridgeon:
> These tests should pass with #279

and they do! Also Laura is using this fix in the wild (the dockerfile fix)

rnpridgeon (approved):

PR #275, opened 2022-04-29
chore(ci): parse repo and branch from github context

arbourd:
Adds a GitHub script (which is just Javascript with the GAPI authenticated) that collects the repository (fork) and branch of the commit that triggered the event.

This is to allow `subo` to checkout forks in CI.

`branch` is either a branch name or tag name.

/cc @yashikajotwani12

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,9 @@ jobs:
 
   test:
     runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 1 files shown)


arbourd:
test on `main`: https://github.com/arbourd/dotfiles/runs/6222667709?check_suite_focus=true
test on a tag: https://github.com/arbourd/dotfiles/runs/6222678471?check_suite_focus=true

javorszky (approved):

PR #276, opened 2022-04-29
chore(builder): update tinygo-base to v0.23.0

arbourd:
To be followed up with a PR to tinygo builder itself

### DIFF
--- a/builder/docker/tinygo/smoke.sh
+++ b/builder/docker/tinygo/smoke.sh
@@ -8,6 +8,7 @@
 
 base=builder/docker/tinygo
 version=$(cat ${base}/.tinygo-ver | tr -d '\n')
+base_version=$(yq e '.substitutions._TINYGO_VERSION' ${base}/cloudbuild.yaml)
 
 err_sum=0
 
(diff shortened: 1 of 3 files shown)


builder/docker/tinygo/smoke.sh
```diff
@@ -44,7 +44,6 @@ if [ $? -ne 0 ]; then
 fi
 
 # Check the base image
-assert_file $base/Dockerfile.base $version 'branch \K.+(?= https://github.com/tinygo-org/tinygo.git)'
```
arbourd, line 47: Removed this line @jagger27 as it is impossible for both the base image and builder image to be updated at the same time.

builder/docker/tinygo/smoke.sh
```diff
@@ -44,7 +45,7 @@ if [ $? -ne 0 ]; then
...
+assert_file $base/Dockerfile.base $base_version 'branch \K.+(?= https://github.com/tinygo-org/tinygo.git)'
```
arbourd, line 48: Updated this line @jagger27 as it is impossible for both the base image and builder image to be updated at the same time.

Instead, it uses `yq` to ensure that the `cloudbuild.yaml` version and the clone are the same.

PR #277, opened 2022-04-29
chore(ci): use GCP CloudBuild for tinygo-base

arbourd:
Use GCP CloudBuild to build Docker `tinygo-base` Docker images. The triggers have already been configured on the GCP side.

- On a pull-request, the images will be built for `arm64` and `amd64` as a test.
- A manual invocation is also available from within the GCP console where a specific commit or tag can be supplied to both build and push the images.

Why GCP? 24h max build runtimes and big machines to hopefully build this faster. CircleCI and GitHub Actions both limit to 5/6h respectively. 

In the future if the tinygo project themselves builds `arm64` images, we can jettison all of this.


builder/docker/tinygo/.tinygo-ver
```diff
@@ -1 +1 @@
-v0.22.0
\ No newline at end of file
+v0.22.0
```
cohix, line 1: Shouldn't this be v0.23.0?

arbourd: This PR is only for whats in main `v0.22.0`. That green line is just my auto formatting adding a newline haha. 

There's a seperate PR for `v0.23.0` which doesn't build yet 😅

cohix: Gotcha :P Proceed

cohix (approved):
:rubberstamp:

PR #278, opened 2022-04-29
chore: simplify Dockerfile

arbourd:
Uses multistage build to simplify the Dockerfile. Places the binary in `/go/bin/subo` as thats where the builder images expect it to be.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,9 @@
-FROM golang:1.18
-
-RUN mkdir -p /root/github.com/suborbital/subo
+FROM golang:1.18-bullseye AS builder
 WORKDIR /root/github.com/suborbital/subo
 
-# dependencies first
 COPY go.* ./
 RUN go mod download
 
-# then everything else
 COPY subo ./subo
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,22 +1,20 @@
...
 COPY project ./project
 COPY scn ./scn
-
 COPY *.go ./
 COPY Makefile .
-
 RUN make subo/docker-bin
+
+FROM debian:bullseye
+COPY --from=builder /go/bin/subo /go/bin/subo
```
arbourd, line 19: Builder PRs expect `subo` to be here.

Dockerfile
```diff
@@ -1,22 +1,20 @@
...
+COPY --from=builder /go/bin/subo /usr/local/bin/subo
```
arbourd, line 20: Should we also put it on the path? We could link too.

Ideally we wanna clean up all these Dockerfiles and not using linking or rely on `$PATH`.

jagger27 (approved):
Looking forward to the smaller images!

PR #280, opened 2022-04-29
chore: simplify builder Dockerfiles

arbourd:
Removes uneeded WORKDIRs. Cleans up order of operations. Mixes some
layers intentionally.

### DIFF
--- a/builder/docker/assemblyscript/Dockerfile
+++ b/builder/docker/assemblyscript/Dockerfile
@@ -1,9 +1,6 @@
 FROM suborbital/subo:dev as subo
 
 FROM node:16-buster-slim
-
-RUN npm install -g npm@latest
-
-COPY --from=subo /go/bin/subo /usr/local/bin
-
 WORKDIR /root/runnable
+COPY --from=subo /go/bin/subo /usr/local/bin
+RUN npm install -g npm@latest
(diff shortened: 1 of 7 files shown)


builder/docker/assemblyscript/Dockerfile
```diff
@@ -1,9 +1,6 @@
...
 FROM node:16-buster-slim
-
-RUN npm install -g npm@latest
-
+WORKDIR /root/runnable
 COPY --from=subo /go/bin/subo /usr/local/bin
-
-WORKDIR /root/runnable
\ No newline at end of file
+RUN npm install -g npm@latest
```
jagger27, line 6: Just for my own curiosity, was moving this line below the `COPY` significant for some weird Docker caching/optimization reason?

builder/docker/grain/Dockerfile
```diff
@@ -1,10 +1,6 @@
...
-WORKDIR /root
-RUN mkdir runnable; mkdir suborbital
-
-COPY --from=subo /go/bin/subo /usr/local/bin
-
-WORKDIR /root/runnable
\ No newline at end of file
+WORKDIR /root/runnable
+COPY --from=subo /go/bin/subo /usr/local/bin/subo
+RUN mkdir mkdir /root/suborbital
```
jagger27, line 6:
```suggestion
RUN mkdir /root/suborbital
```
Extra `mkdir`

jagger27 (commented):
I see that we removed `mkdir runnable` everywhere, but we're still using `WORKDIR /root/runnable`. Where is that folder being created now?

builder/docker/assemblyscript/Dockerfile
arbourd: In this case it shouldn't really matter, but ya it's always best to have runs at the bottom to maintain prevent cache busting.

jagger27 (approved):

PR #284, opened 2022-05-04
feat: update tinygo to v0.23.0

arbourd:


jagger27 (approved):
We’ll have to see how this works with Reactr’s test suite, but that hasn’t been an issue in the past.

PR #302, opened 2022-05-16
chore: release binary after docker images

arbourd:
Subo requires Docker images to build runnables. This change ensures that
all images are built and pushed before building and pushing the binary.


tarynmusgrave (approved):
👍🏻

PR #337, opened 2022-08-01
fix(builder): Allow JS to bind to resp_set_header

ospencer:
cc @flaki


flaki (approved):
This looks great, swapped out the custom build and managed to use this with functions.fly.dev and co. After the runtime bits land and go on npm, I can check the whole end-to-end workflow.

PR #342, opened 2022-08-04
Flaki/175 fix root

### LINKED ISSUE #175: Docker toolchain build artifacts are created under the root user on Linux
I'm seeing a weird issue on Linux (Ubuntu 20.04, running as my own user `flaki`) where `subo build .` would use the docker build method to build the Rust runnables, and while the `runnables.wasm.zip` bundle comes out perfectly fine (`flaki:flaki` user/group), the build artifacts end up in folders that are `root:root`-owned and cannot be deleted without using `sudo`.

image

As a result, `subo clean` fails also without a `sudo`:
image

flaki:
Fixes #175


flaki:
Updated to track `main` instead of `vmain`

flaki:
The latest update:
- Pulls in changes from #355, this allows us to use `subo/toolchaintest` (an updated `smoketest`) to build & test the builder Docker toolchain images locally
- Adds a `--verbose` flag to `subo build` which enables logging of the Docker toolchain commands for debugging
- Explicitly sets `HOME=/tmp` when invoking the Docker toolchains

About that last part:

We are calling the Docker toolchain as a limited user, quite possibly a user that does not even exist inside the container (thus, `HOME=/`). This is not a problem as long as we only read world-readable files and only write to world-writable directories. Most toolchains (e.g. Swift, Go) will write temporary files (like `.cache`-s) to the `HOME` folder and will break when attempting to put files into `/`, setting `HOME` to `/tmp` resolves this issue, and since these are temporary files we don't need them anyway.

`toolchaintest` is passing for me on Linux but it would be good to have someone have a look at this on OSX and see how it fares.

root.go
```diff
@@ -26,8 +26,8 @@ including building WebAssembly Runnables and Atmo projects.`,
 	// create commands.
 	create := &cobra.Command{
 		Use:   "create",
-		Short: "create a runnable, project, or handler",
-		Long:  `create a new Atmo project, WebAssembly runnable or handler`,
+		Short: "create a module, project, or handler",
+		Long:  `create a new DeltaV project, WebAssembly module or handler`,
```
ospencer, line 30:
```suggestion
		Long:  `create a new E2Core project, WebAssembly module or handler`,
```

ospencer (commented):
This is sweet! Someone with more Docker expertise than me should probably be the one to approve though :)

builder/builder.go
```diff
@@ -157,7 +157,18 @@ func (b *Builder) dockerBuildForLang(lang string) (*BuildResult, error) {
 
 	result := &BuildResult{}
 
-	outputLog, err := b.Config.CommandRunner.Run(fmt.Sprintf("docker run --rm --mount type=bind,source=%s,target=/root/runnable %s subo build %s --native --langs %s", b.Context.MountPath, img, b.Context.RelDockerPath, lang))
+	// Run the Docker toolchain as the current user
+	uid := os.Getuid()
+	gid := os.Getgid()
+
+	toolchainCmd := fmt.Sprintf("docker run --rm --mount type=bind,source=%s,target=/usr/src/runnable -u %d:%d --env HOME=/tmp %s subo build %s --native --langs %s", b.Context.MountPath, uid, gid, img, b.Context.RelDockerPath, lang)
```
jagger27, line 164: This is great and obvious in hindsight!

builder/docker/javascript/Dockerfile
```diff
@@ -2,11 +2,11 @@ FROM ghcr.io/suborbital/javy:v0.3.0 as javy
 FROM suborbital/subo:dev as subo
 
 FROM node:16-bullseye-slim
-WORKDIR /root/runnable
+WORKDIR /usr/src/runnable
 # Propagate our root permissions for our home folder to everyone. This allows
 # npm scripts (which get run as whatever user owns the mounted runnable
 # directory) to access common home folder resources (caches, etc.).
-RUN mkdir /root/suborbital && chmod -R o=u /root
+# RUN mkdir /root/suborbital && chmod -R o=u /root
```
jagger27, line 9:
```suggestion
```

builder/docker/assemblyscript/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM suborbital/subo:dev as subo
 
 FROM node:16-buster-slim
-WORKDIR /root/runnable
+WORKDIR /usr/src/runnable
```
jagger27, line 4: Why was this path chosen? `/usr/*` paths are usually root owned as well and I don't see a `chown` anywhere for it. Does all of the writing happen in the `/tmp` dir set as `$HOME` in the Docker invocation?

scripts/toolchaintest.sh
```diff
@@ -0,0 +1,53 @@
...
+fi
+
+# rebuild local docker build tooling
+docker build . -f builder/docker/assemblyscript/Dockerfile -t suborbital/builder-as:dev
+docker build . -f builder/docker/grain/Dockerfile --platform linux/amd64 -t suborbital/builder-gr:dev
+docker build . -f builder/docker/javascript/Dockerfile -t suborbital/builder-js:dev
+docker build . -f builder/docker/rust/Dockerfile -t suborbital/builder-rs:dev
+docker build . -f builder/docker/swift/Dockerfile -t suborbital/builder-swift:dev
+docker build . -f builder/docker/tinygo/Dockerfile --platform linux/amd64 --build-arg TARGETARCH=amd64 -t suborbital/builder-tinygo:dev
+docker build . -f builder/docker/wat/Dockerfile -t suborbital/builder-wat:dev
```
jagger27, line 36: We don't have an explicit `dev` target for local builds (which is weird) but these can be replaced with calls to make like this:

```bash
make ver=dev builder/docker/tinygo
```

builder/docker/javascript/Dockerfile
flaki: Thanks, I didn't want to touch it until we knew it was harmless to remove. :)

builder/docker/assemblyscript/Dockerfile
flaki: I...honestly don't remember where `/usr/src` came from, I definitely remember `/root` being an issue though.

The `/tmp` as HOME change came later for the Go/Swift images and I don't remember that being needed for other images, but I think this might vary depending on what's the UID/GID of the invoking user.

scripts/toolchaintest.sh
flaki: Cool, I'll change it around!

Issue #327, opened 2022-07-11
Subo should give a more descriptive output for `create token`


ospencer:
Right now, Subo just prints out the token. It would be helpful if it said something along the lines of "here is your environment token, keep it safe".

PR #349, opened 2022-09-07
docs: add context before printing token

LauraLangdon:
Closes Issue #327.

### DIFF
--- a/subo/command/compute_create_token.go
+++ b/subo/command/compute_create_token.go
@@ -50,7 +50,7 @@ func ComputeCreateTokenCommand() *cobra.Command {
 func getVerifierCode() (string, error) {
 	fmt.Print("A verification code was sent to your email address. " +
 		"Enter the code to continue, " +
-		"and your token will print below (keep it safe!): ")
+		"and your environment token will print below (keep it safe!): ")
 	code, err := input.ReadStdinString()
 	if err != nil {
 		return "", errors.Wrap(err, "failed to ReadStdinString")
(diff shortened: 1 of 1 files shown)


javorszky (approved):
Nice use of breaking up the string to keep lines in editor short! 🎉 

Is the expectation that this would print in 3 lines on the terminal as well? 🙂

subo/command/compute_create_token.go
```diff
@@ -48,7 +48,9 @@ func ComputeCreateTokenCommand() *cobra.Command {
 
 // getVerifierCode gets the 6-character code from stdin.
 func getVerifierCode() (string, error) {
-	fmt.Print("A verification code was sent to your email address. Enter the code to continue: ")
+	fmt.Print("A verification code was sent to your email address. " +
+		"Enter the code to continue, " +
+		"and your token will print below (keep it safe!): ")
```
cohix, line 53: Perhaps we should be specific and say 'your environment token' so that it exactly matches the nomenclature we use in the docs?

cohix: Also, you can use backticks instead of double-quotes to do multi-line strings in Go, if you wish to do so :)

LauraLangdon: > Perhaps we should be specific and say 'your environment token' so that it exactly matches the nomenclature we use in the docs?

Done!

LauraLangdon: > Also, you can use backticks instead of double-quotes to do multi-line strings in Go, if you wish to do so :)

Thanks! Is there a standard around that for us?

cohix: Nope! It's a Go language feature. If you want it to be one line, use double quotes, if you want it to be multi-line, use backticks.

cohix (approved):

PR #351, opened 2022-09-08
docs: add instructions for upgrading Subo

### LINKED ISSUE #350: Provide instructions for upgrading Subo
Currently Subo alerts users when there's an upgrade available for Subo, but doesn't provide any guidance as to how to upgrade. This is complex, because the method of upgrade depends on the method of installation, and we don't yet have a means of detecting which method was used (tracked in https://github.com/suborbital/subo/issues/147). So we can work on this in two phases, one now and one when we can solve the installation method problem. This issue will address Phase 1.

- Phase 1: Link to:
  - [ ] install page to show the possible methods that could have been used to install Subo originally
  - [ ] Discord for help

LauraLangdon:
Closes #350.

### DIFF
--- a/subo/release/check.go
+++ b/subo/release/check.go
@@ -11,7 +11,6 @@ import (
 	"time"
 
 	"github.com/google/go-github/v41/github"
-	"github.com/hashicorp/go-version"
 	"github.com/pkg/errors"
 
 	"github.com/suborbital/subo/subo/util"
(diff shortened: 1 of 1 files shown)


subo/release/check.go
```diff
@@ -11,7 +11,6 @@ import (
 	"time"
 
 	"github.com/google/go-github/v41/github"
-	"github.com/hashicorp/go-version"
```
rnpridgeon, line 14: I think you still rely on this dependency

subo/release/check.go
```diff
@@ -11,7 +11,6 @@ import (
 	"time"
 
 	"github.com/google/go-github/v41/github"
```
rnpridgeon, line 13:
```suggestion
	"github.com/google/go-github/v41/github"
	"github.com/hashicorp/go-version"
```

javorszky (approved):

rnpridgeon (approved):

PR #353, opened 2022-09-09
fix: Use `docker compose`, not `docker-compose`

### LINKED ISSUE #344: Subo should support systems with docker compose (v2)
I installed Docker [from the repository](https://docs.docker.com/engine/install/ubuntu/#install-using-the-repository) on Ubuntu, side-stepping Docker Desktop, and so I only have `docker-compose-plugin` which is the new `docker compose` (note the space).

> The docker compose (with a space) is a newer project to migrate compose to Go with the rest of the docker project. This is the v2 branch of the [docker/compose](https://github.com/docker/compose/tree/v2) repo. It's been first introduced to Docker Desktop users, so docker users on Linux didn't see the command. In addition to migrating to Go, it uses the compose-spec, and part of the rewrite may result in behavior differences.

Apparently [this is the way of the future](https://stackoverflow.com/a/66526176), and `subo compute deploy core --local` currently fails with an error, not being able to find the old `docker-compose` command:

```
⏩ START: preparing deployment
ℹ️  using cached environment token
✅ DONE: ready to start installation
⏩ START: installing...
sh: 1: docker-compose: not found
Error: 🚫 failed to docker-compose up: failed to Run command: exit status 127
Usage:
  subo compute deploy core [flags]
```

Because `alias`-es don't work with subo, one actually needs to create a [shell shim](https://askubuntu.com/a/955919) that re-maps `docker-compose` to `docker compose` (and that works because `up` is the same in both CLI-s), but this is a hassle and is undocumented, so we should at least document it, but it would be better fixing it in `subo` itself.

callahad:
Compose v2 is the future, and the legacy standalone docker-compose script may not be present on all installs.

Fixes #344

### DIFF
--- a/subo/command/compute_deploy_core.go
+++ b/subo/command/compute_deploy_core.go
@@ -129,11 +129,12 @@ func ComputeDeployCoreCommand() *cobra.Command {
 			util.LogStart("installing...")
 
 			if localInstall {
-				if _, err := util.Command.Run("docker-compose up -d"); err != nil {
-					return errors.Wrap(err, "🚫 failed to docker-compose up")
+				if _, err := util.Command.Run("docker compose up -d"); err != nil {
+					util.LogInfo("Is Docker Compose installed? https://docs.docker.com/compose/install/")
+					return errors.Wrap(err, "🚫 failed to run `docker compose up`")
 				}
 
-				util.LogInfo("use `docker ps` and `docker-compose logs` to check deployment status")
+				util.LogInfo("use `docker ps` and `docker compose logs` to check deployment status")
 
 				proxyPortStr := strconv.Itoa(proxyPort)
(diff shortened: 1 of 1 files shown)


callahad:
See #354 for an alternative which also falls back to `docker-compose` if Compose v2 is not detected.

callahad:
(I suspect my Go may not be idiomatic here.)

The Compose v2 plugin should be present on all Docker Desktop installations updated in the last 18 months. It's most likely to be missing for Linux users who installed the Docker engine directly and accidentally missed the separate `docker-compose-plugin` package. The right answer is for them to install the plugin... but some proportion of those folks are likely to still have the legacy `docker-compose` script on their systems, and this PR would make things Just Work for them.

callahad:
Closing in favor of #354

PR #354, opened 2022-09-09
fix: Prefer `docker compose` to `docker-compose`

### LINKED ISSUE #344: Subo should support systems with docker compose (v2)

callahad:
Compose v2 is the future, but it may not be present on all installs.

This PR prefers `docker compose` but falls back to `docker-compose` if Compose v2 is not detected.

Alternative to #353

Fixes #344

### DIFF
--- a/subo/command/compute_deploy_core.go
+++ b/subo/command/compute_deploy_core.go
@@ -129,11 +129,12 @@ func ComputeDeployCoreCommand() *cobra.Command {
 			util.LogStart("installing...")
 
 			if localInstall {
-				if _, err := util.Command.Run("docker-compose up -d"); err != nil {
-					return errors.Wrap(err, "🚫 failed to docker-compose up")
+				if _, err := util.Command.Run("docker compose up -d"); err != nil {
+					util.LogInfo("Is Docker Compose installed? https://docs.docker.com/compose/install/")
+					return errors.Wrap(err, "🚫 failed to run `docker compose up`")
 				}
 
-				util.LogInfo("use `docker ps` and `docker-compose logs` to check deployment status")
+				util.LogInfo("use `docker ps` and `docker compose logs` to check deployment status")
 
 				proxyPortStr := strconv.Itoa(proxyPort)
 				proxy := localproxy.New("editor.suborbital.network", proxyPortStr)
@@ -179,7 +180,7 @@ func ComputeDeployCoreCommand() *cobra.Command {
 	cmd.Flags().String(branchFlag, "main", "git branch to download templates from")
 	cmd.Flags().String(versionFlag, release.SCCTag, "Docker tag to use for control plane images")
 	cmd.Flags().Int(proxyPortFlag, proxyDefaultPort, "port that the Editor proxy listens on")
-	cmd.Flags().Bool(localFlag, false, "deploy locally using docker-compose")
+	cmd.Flags().Bool(localFlag, false, "deploy locally using docker compose")
 	cmd.Flags().Bool(dryRunFlag, false, "prepare the deployment in the .suborbital directory, but do not apply it")
 	cmd.Flags().Bool(resetFlag, false, "reset the deployment to default (replaces docker-compose.yaml and/or Kubernetes manifests)")
 
--- a/subo/command/compute_deploy_core.go
+++ b/subo/command/compute_deploy_core.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


callahad:
(I suspect my Go may not be idiomatic here.)

The Compose v2 plugin should be present on all Docker Desktop installations updated in the last 18 months. It's most likely to be missing for Linux users who installed the Docker engine directly and accidentally missed the separate `docker-compose-plugin` package. The right answer is for them to install the plugin... but some proportion of those folks are likely to still have the legacy `docker-compose` script on their systems, and this PR would make things Just Work for them.

...but it's much simpler to just use Compose v2 directly and encourage people to install the plugin if it fails.

subo/command/compute_deploy_core.go
```diff
@@ -129,11 +130,27 @@ func ComputeDeployCoreCommand() *cobra.Command {
...
+					// See: https://github.com/containerd/nerdctl/issues/1368
+					compose = "docker compose"
 				}
 
-				util.LogInfo("use `docker ps` and `docker-compose logs` to check deployment status")
+				command := fmt.Sprintf("%s up -d", compose)
+
+				if _, err := util.Command.Run(command); err != nil {
+					util.LogInfo("Is Docker Compose installed? https://docs.docker.com/compose/install/")
+					return errors.Wrap(err, "🚫 failed to run `docker compose up`")
```
cohix, line 150: I would put the _actual_ command that was run in the error message

flaki:
LGTM, would def prefer this to #353

subo/command/compute_deploy_core.go
```diff
@@ -129,11 +130,27 @@ func ComputeDeployCoreCommand() *cobra.Command {
...
 
-				util.LogInfo("use `docker ps` and `docker-compose logs` to check deployment status")
+				command := fmt.Sprintf("%s up -d", compose)
+
+				if _, err := util.Command.Run(command); err != nil {
+					util.LogInfo("Is Docker Compose installed? https://docs.docker.com/compose/install/")
+					return errors.Wrapf(err, "🚫 failed to run `%s`", command)
+				}
+
+				util.LogInfo("use `docker ps` and `docker compose logs` to check deployment status")
```
javorszky, line 153: this line should also use whatever the `compose` variable is.
```suggestion
				util.LogInfo(fmt.Sprintf("use `docker ps` and `%s logs` to check deployment status", compose))
```

subo/command/compute_deploy_core.go
```diff
@@ -179,7 +196,7 @@ func ComputeDeployCoreCommand() *cobra.Command {
 	cmd.Flags().String(branchFlag, "main", "git branch to download templates from")
 	cmd.Flags().String(versionFlag, release.SCCTag, "Docker tag to use for control plane images")
 	cmd.Flags().Int(proxyPortFlag, proxyDefaultPort, "port that the Editor proxy listens on")
-	cmd.Flags().Bool(localFlag, false, "deploy locally using docker-compose")
+	cmd.Flags().Bool(localFlag, false, "deploy locally using docker compose")
```
javorszky, line 199: Should also maybe use whatever the code decided above docker-compose to be? Though that might mean that we might need to move the code to determine whether the system has docker-compose or docker compose available to be outside / before the `if localInstall` branch.

javorszky (changes requested):
I think some additional refinements can be made with logging to user if we're determining which one to use.

subo/command/compute_deploy_core.go
callahad: Or I could just use the proper noun version 😉 

```suggestion
	cmd.Flags().Bool(localFlag, false, "deploy locally using Docker Compose")
```

callahad: This is such a niche case I'm inclined to actually do that

callahad: ...and there's precedent up on line 39:

    Long:  `deploy the Suborbital Compute Core using Kubernetes or Docker Compose`

javorszky: Which might work. Is there a way to make this super clear to our users? I'm thinking about a user who may be an experienced developer, but new to docker. Would they know the difference between `docker compose` (the command in the terminal) and `Docker Compose` (the thing they're invoking with the above command(s))?

Is there a way to make this clear / remove ambiguity?

subo/command/compute_deploy_core.go
callahad: Incorporated into 2413626bbb7b42a420943d6aaddb0ffcac23dfeb

javorszky (approved):
All good, we talked about the decisions behind all parts of it. 🚢

subo/command/compute_deploy_core.go
callahad: Inclined to leave as-is: for most users, there is no distinction (`docker compose` is present on all Docker Desktop installs updated in the last year *and* `docker-compose` is shimmed to work like an alias to `docker compose`). Pointing out the historic separation gets nuanced enough that it may actually create confusion 😆

callahad:
🙏 Thank you very much for the reviews

PR #355, opened 2022-09-13
Update smoketest commands to replace subo create runnable

hola-soy-milk:
I noticed that `subo create runnable` no longer works. @cohix am I correct in assuming it should be `subo create module` now?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,9 @@ subo/docker/publish:
 subo/smoketest: subo
 	./scripts/smoketest.sh
 
+subo/toolchaintest: subo/docker
+	./scripts/toolchaintest.sh
+
 mod/replace/atmo:
 	go mod edit -replace github.com/suborbital/atmo=$(HOME)/Workspaces/suborbital/atmo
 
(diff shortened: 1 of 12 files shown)


flaki:
So I rebased this upon #360 but somoketests will still fail, simply because we have updated the module (née "runnable") detection code from `.runnable.yaml`-s to `.module.yaml`-s, which change needs to go into `subo` first, then we need to update the `builder` Docker images because they bundle subo -- and you guessed it, the smoketest uses the Docker toolchain.

flaki:
This last commit adds `make subo/toolchaintest` via `./scripts/toolchaintest.sh` that is largely the same as the smoketest except it builds its own local `dev` images and uses them. I yoinked out these image builds from `builder.mk`, though they are largely the same as the builds in there so in the future we could refactor this and get rid of the duplication but for now this means that `make subo/toolchaintest` will build new images and test project & module creation; currently this test should be passing (it does on my machine).

arbourd:
Closing this as we recently implemented this (and split the templates). Thanks @hola-soy-milk!

PR #358, opened 2022-09-14
fix: add mising files

cohix:


subo/command/compute_deploy_core.go
```diff
@@ -193,7 +193,7 @@ func ComputeDeployCoreCommand() *cobra.Command {
 		},
 	}
 
-	cmd.Flags().String(branchFlag, "main", "git branch to download templates from")
+	cmd.Flags().String(branchFlag, defaultBranch, "git branch to download templates from")
```
rnpridgeon, line 196: I may have missed it skimming but I don't see where defaultBranch is defined.

rnpridgeon: nvm found it, 

https://github.com/suborbital/subo/blob/3bdb983c9f8b253eb6356f56cad1f1ee628368e2/subo/command/create_project.go#L18

rnpridgeon (approved):

subo/command/compute_deploy_core.go
cohix: That was in a previous PR

PR #359, opened 2022-09-15
fix: clarify option selection instruction in REPL

### LINKED ISSUE #352: Expand option selection instruction in the REPL
Currently users are asked to choose an option, where the only option is "Create/edit" a function. We'd like to split these into two separate options in the future, but for this moment we'll just clarify the wording.

LauraLangdon:
Closes #352


LauraLangdon:
Closing since Subo's being deprecated.

PR #362, opened 2022-09-21
fix: remove user-facing Atmo mentions

### LINKED ISSUE #361: Remove Atmo mentions
@flaki noticed that Subo mentions Atmo to users when it [lists available commands](https://github.com/suborbital/subo/blob/3bdb983c9f8b253eb6356f56cad1f1ee628368e2/subo/command/dev.go) and [at the root](https://github.com/suborbital/subo/blob/3bdb983c9f8b253eb6356f56cad1f1ee628368e2/root.go). We can either:

- Do a 1:1 name change
- Remove the reference entirely

The second option seems to me to be most in line with our new positioning, so I'm starting with that approach. 😊

LauraLangdon:
Closes #361. I just removed the mentions to Atmo rather than replacing them, but could do the opposite if that would be preferred ofc!

### DIFF
--- a/root.go
+++ b/root.go
@@ -13,8 +13,7 @@ func rootCommand() *cobra.Command {
 		Use:     "subo",
 		Short:   "Suborbital Development Platform CLI",
 		Version: release.Version(),
-		Long: `Subo is the full toolchain for using and managing Suborbital
-Development Platform tools.`,
+		Long:    `Subo is the full toolchain for using and managing Suborbital Extension Engine tools.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cmd.Help()
 			return nil
--- a/root.go
+++ b/root.go
@@ -13,8 +13,8 @@ func rootCommand() *cobra.Command {
 		Use:     "subo",
 		Short:   "Suborbital Development Platform CLI",
 		Version: release.Version(),
-		Long: `Subo is the full toolchain for using and managing Suborbital Development Platform tools,
-including building WebAssembly Runnables and Atmo projects.`,
+		Long: `Subo is the full toolchain for using and managing Suborbital
+Development Platform tools.`,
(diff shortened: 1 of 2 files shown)


root.go
```diff
@@ -13,8 +13,8 @@ func rootCommand() *cobra.Command {
 		Use:     "subo",
 		Short:   "Suborbital Development Platform CLI",
 		Version: release.Version(),
-		Long: `Subo is the full toolchain for using and managing Suborbital Development Platform tools,
-including building WebAssembly Runnables and Atmo projects.`,
+		Long: `Subo is the full toolchain for using and managing Suborbital
+Development Platform tools.`,
```
cohix, line 17: I don't think we need the linebreak here, and 'Suborbital Development Platform' can be replaced with 'Suborbital Extension Engine'

subo/command/dev.go
```diff
@@ -16,8 +16,8 @@ import (
 func DevCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "dev",
-		Short: "run a development Atmo server using Docker",
-		Long:  `run a development Atmo server using Docker`,
+		Short: "run a development  server using Docker",
```
cohix, line 19: I think there's an extra space here

subo/command/dev.go
```diff
@@ -54,8 +54,9 @@ func DevCmd() *cobra.Command {
 		},
 	}
 
-	cmd.Flags().String("port", "8080", "set the port that Atmo serves on")
-	cmd.Flags().BoolP("verbose", "v", false, "run Atmo with debug level logging")
+	cmd.Flags().String("port", "8080",
+		"set the port on which to serve the project")
```
cohix, line 58: extra linebreak?

ospencer (approved):

PR #371, opened 2022-11-15
chore: Use renovate to update tiny-go

arbourd:
Uses a Regex string and manager to capture the version and update it.

Example of the regex at: https://regex101.com/r/lfZC3R/1

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,7 +1,20 @@
 {
-  enabledManagers: ["github-actions", "gomod"],
-  postUpdateOptions: [
-    "gomodTidy", // go mod tidy
-    "gomodUpdateImportPaths",
+  enabledManagers: ["github-actions", "gomod", "regex"],
+  postUpdateOptions: ["gomodTidy", "gomodUpdateImportPaths"],
+  regexManagers: [
+    {
+      fileMatch: ["Dockerfile$"],
+      matchStrings: [
+        "datasource=(?<datasource>.*?) depName=(?<depName>.*?)( versioning=(?<versioning>.*?))?\\n.*_VERSION=(?<currentValue>.*)",
+      ],
+      versioningTemplate: "{{#if versioning}}{{{versioning}}}{{else}}semver{{/if}}",
+    },
+  ],
+  packageRules: [
+    {
+      packagePatterns: [".*"],
+      managers: ["regex"],
+      extractVersion: "^(v)?(?<version>.*)$",
+    },
   ],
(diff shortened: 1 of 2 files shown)


.github/renovate.json5
```diff
@@ -1,7 +1,20 @@
 {
-  enabledManagers: ["github-actions", "gomod"],
-  postUpdateOptions: [
-    "gomodTidy", // go mod tidy
-    "gomodUpdateImportPaths",
+  enabledManagers: ["github-actions", "gomod", "regex"],
+  postUpdateOptions: ["gomodTidy", "gomodUpdateImportPaths"],
+  regexManagers: [
+    {
+      fileMatch: ["Dockerfile$"],
```
javorszky, line 6: this will only catch files where `Dockerfile` is at the end of the file name. Do we want to restrict it to those?

```
something-Dockerfile // will match
Dockerfile // will match
Dockerfile-gci // no match
```

.github/renovate.json5
```diff
@@ -1,7 +1,20 @@
...
+      matchStrings: [
+        "datasource=(?<datasource>.*?) depName=(?<depName>.*?)( versioning=(?<versioning>.*?))?\\n.*_VERSION=(?<currentValue>.*)",
```
javorszky, line 8: I see this relates to line 9 of the Dockerfile. I assume the point is to keep the version number on line 10 up to date. How will that be achieved? I'm having trouble figuring out what part of this config will say "update the `ARG TINYGO_VERSION=<version>` to whatever the latest version of `tinygo-org/tinygo` is"

.github/renovate.json5
arbourd: The goal is just to catch the subo Dockerfiles so that any `*_VERSION=` string could be auto updated.

.github/renovate.json5
arbourd: It's kinda magical, but it uses [this](https://docs.renovatebot.com/modules/manager/regex/). It depends on the capture group `currentValue` to determine what the version is and replaces it, if the remote value is larger semver.

javorszky: right, I see. So it checks for two lines. One is this one:
```
# renovate: datasource=github-releases depName=tinygo-org/tinygo
```
and the other line is one immediately following that which has a `_VERSION=semver` in it somewhere, and it's kinda strict that these two go together, and that the line above relates to the version below

javorszky (approved):

PR #380, opened 2022-11-17
chore: Use latest golangci-lint

arbourd:


.github/workflows/ci.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3
         with:
-          version: v1.45
+          version: latest
```
javorszky, line 25: I'm against setting this to a floating version. I'd rather periodically pin it to current latest at the time.

Reason being if golangci-lint updates to a version that breaks things, we'll see all our tests fail for subo due to a thing that's not in our control.

arbourd: I understand that it's not atomic, but it's also annoying to constantly update minors and I'm more concerned about us (or say a contributor) getting stuck on a 2 year old linter. Ideally, we'd probably want "1.x" but I don't remember that being an option.

I think there's value in getting constant feedback from the linter (for better or worse) rather than expecting someone to update this. 

Additionally, it's consistent across all our other repositories.

javorszky: All right, let's run with it, and fix it if / when we run into issues

javorszky (approved):

PR #382, opened 2022-11-17
chore: Escape strings for title

arbourd:
Prevents the title from throwing a stat error if there is a space in it.


arbourd:
Maybe we actually want this to fail, probably won't keep this release structure for ever anyways so why fix/maintain it

PR #385, opened 2022-11-21
chore: Update subo to use SE2 images over SCC/SCN

arbourd:
Moves over to SE2 template names. Updates SE2 to `v0.4.1`.

### DIFF
--- a/subo/command/se2_deploy.go
+++ b/subo/command/se2_deploy.go
@@ -23,7 +23,7 @@ import (
 )
 
 type deployData struct {
-	SCCVersion       string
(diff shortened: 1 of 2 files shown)


cohix:
@arbourd we likely need to update the `defaultRepo` and `defaultBranch` consts to point to `suborbital/templates` and `main` as I believe it's currently `suborbital/runnable-templates` and `vmain`.

arbourd:
@cohix That was changed in 383. Tests rely upon this working, methinks.

cohix (approved):

PR #387, opened 2022-11-21
chore: Remove references to Atmo, Reactr and Runnables

arbourd:
Replaced with E2Core and modules.


scn/types/heartbeat.go
```diff
@@ -2,12 +2,12 @@ package types
 
 // HeartbeatRequest is a request to send heartbeat data.
 type HeartbeatRequest struct {
-	Version   string         `json:"version"`
-	Runnables *RunnableStats `json:"runnables"`
```
arbourd, line 6: What uses this?

cohix: That's the payload that the SE2 API service receives when telemetry is sent. I don't think we can change this without it breaking telemetry.

cohix (approved):
A few small things

PR #390, opened 2022-11-24
fix: deploy the e2core version from subo, not the templates

arbourd:
E2Core version is currently hardcoded into the templates (suborbital/templates) repository directly. This commit templates the release.RuntimeVersion (which represents E2Core) into the templates, rather than rely on the hardcoded tag in the template itself.

This commit also makes some small changes that:
- Rename FFIVersion to SDKVersion (as Reactr will be SDK)
- Set release vars as consts
- Differentiates versions and tags

### DIFF
--- a/subo/command/se2_deploy.go
+++ b/subo/command/se2_deploy.go
@@ -23,7 +23,10 @@ import (
 )
 
 type deployData struct {
-	SE2Version       string
+	SE2Version string // TODO: remove me after Beta-6 release
+
+	SE2Tag           string
+	E2CoreTag        string
 	EnvToken         string
 	BuilderDomain    string
 	StorageClassName string
@@ -93,7 +96,9 @@ func SE2DeployCommand() *cobra.Command {
 				}
 
 				data := deployData{
+					SE2Tag:     tag,
 					SE2Version: tag,
+					E2CoreTag:  "v" + release.RuntimeVersion,
 					EnvToken:   envToken,
 				}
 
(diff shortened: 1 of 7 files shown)


subo/command/se2_deploy.go
```diff
@@ -93,7 +96,9 @@ func SE2DeployCommand() *cobra.Command {
 				}
 
 				data := deployData{
+					SE2Tag:     tag,
 					SE2Version: tag,
```
arbourd, line 100: Prefer `XTag` here because they represent tags, not versions (which are similar but different because of prefixed v).

subo/command/se2_deploy.go
```diff
@@ -23,7 +23,10 @@ import (
 )
 
 type deployData struct {
-	SE2Version       string
+	SE2Version string // TODO: remove me after Beta-6 release
+
+	SE2Tag           string
+	E2CoreTag        string
```
arbourd, line 29: Prefer `XTag` here because they represent tags, not versions (which are similar but different because of prefixed v).

subo/command/se2_deploy.go
```diff
@@ -23,7 +23,10 @@ import (
 )
 
 type deployData struct {
-	SE2Version       string
+	SE2Version string // TODO: remove me after Beta-6 release
```
arbourd, line 26: `SE2Version` is used by templates on main now, but will be renamed to `SE2Tag`

subo/release/version.go
```diff
@@ -1,14 +1,13 @@
...
+// SDKVersion is the SDK version used by this version of subo.
+const SDKVersion = "0.15.1"
 
 // RuntimeVersion is the default version of E2Core that will be used for new projects.
-var RuntimeVersion = "0.4.7"
+const RuntimeVersion = "0.4.7"
 
 // SE2Tag is the docker tag used for creating new SE2 deployments.
-var SE2Tag = "v0.4.1"
+const SE2Version = "0.4.1"
```
arbourd, line 13: For consistency, make this a version and not a tag.

jagger27 (approved):

PR #395, opened 2022-11-30
chore: Rename function to module or extension

arbourd:


subo/repl/repl.go
```diff
@@ -46,7 +46,7 @@ func (r *Repl) Run() error {
 
 		switch opt {
 		case "1":
-			replErr = r.editFunction()
+			replErr = r.editModule()
```
ospencer, line 49: I would call it `editExtension`.

ospencer (approved):

PR #398, opened 2022-12-08
feat: Compute -> E2Core Storage Migration

rnpridgeon:
Testing:

Manually tested against minio (s3 compatible), gcs (thanks dylan) and a temporary directory. 

Setup 
1. Build subo 5.4, start compute creates some functions in the repl shut it down - storage root "./old"
2. Install subo 6, start compute, create some function in the repl shut it down - storage root "./new"
3. Build PR branch exec `se2 migrate [storage root]`
4. Verify contents

In SCN I wrote some adhoc tests using the storage instances and pointed them at my test storage to confirm i could list tenatnts, find tenants, read the system value etc. 

How it works: 

1. Walks storage and marks all assets of interest in a trie
   a. each asset is placed under a category root (tenant, runnable, bin, src) and points to the underlying resource

2. Process each module binary so we can obtain the new ref (sha256 digest), create the new directory (ref) and copy the binary to its home. During this time we record the ref and place it in the tree in a new category (digest). 

3. Next we process the drafts, this requires looking up it's associative ref from the tree then creating a new directory, and finally copying the draft contents into that directory. 

4. Next we process all runnables which are translated into "modules" and recorded at the module rot directory. 

5. Next we need to collect all modules we process and record them in a tenant overview which is then placed at the tenant root directory. 

6. Finally we create a somewhat superficial system version number and place it in the system storage root. This mostly just needs to be greater than 0 so e2core can see that there are modules there for it to discover; subsequent deploys will increment this appropriately from there.

I decided to leave clean up as an exercise for the user to ensure they can gracefully recover in the event something terrible happens which didn't reproduce in my tests.


go.mod
```diff
@@ -1,29 +1,60 @@
 module github.com/suborbital/subo
 
-go 1.19
+go 1.18
```
rnpridgeon, line 3: oops, will fix when I fix the linter tomorrow

subo/util/storage_migrate.go
```diff
@@ -0,0 +1,618 @@
...
+
+var versionPattern = regexp.MustCompile(`v[0-9]*\.0\.0`)
+
+type Tenant struct {
+	Overview *system.TenantOverview `json:"overview,omitempty"`
+	Stale    bool                   `json:"stale"`
+}
+
+// Runnable stores the relevant parts of a .runnable.yaml file
+type Runnable struct {
```
rnpridgeon, line 38: viva la compute!

subo/util/storage_migrate.go
```diff
@@ -0,0 +1,618 @@
...
+	Name         string `yaml:"name" json:"name"`
+	Namespace    string `yaml:"namespace" json:"namespace"`
+	Lang         string `yaml:"lang" json:"lang"`
+	Version      string `yaml:"version" json:"version"`
+	DraftVersion string `yaml:"draftVersion,omitempty" json:"draftVersion,omitempty"`
+	APIVersion   string `yaml:"apiVersion,omitempty" json:"apiVersion,omitempty"`
+	FQFN         string `yaml:"fqfn" json:"fqfn"`
+}
+
+func main() {
```
rnpridgeon, line 49: oops, will delete

subo/util/storage_migrate.go
```diff
@@ -0,0 +1,618 @@
...
+			return nil
+		}
+
+		// map modules
+		if !d.IsDir() && d.Name() == runnable.Name+".wasm" {
+			recorder.Insert(canonicalize(&runnable, "bin", version(path)), path)
+			return nil
+		}
+
+		// map recorder
```
rnpridgeon, line 110: s/recorder/draft

subo/command/se2_migrate_storage.go
```diff
@@ -0,0 +1,45 @@
...
+func SE2MigrateStorageCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "migrate [storage root]",
+		Short: "reformat storage to match SE2 specification",
+		Long:  `Update storage format to match the Suborbital Extension Engine (SE2) storage specification`,
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			location := args[0]
+
+			switch scheme(location) {
```
ospencer, line 22: shouldn't this be the thing with the call to `ToLower`?

rnpridgeon: yes 🤦

ospencer (approved):

